PR #1, opened 2020-05-20
Next.js spike

trcywu:


frigus02:
Agree. Let's merge this in order to get the CI work going.

Before moving the ticket to done, I would like to rename the folder `future-nhs` to something like `web` or `frontend` and write a readme, which fits to us.

PR #2, opened 2020-05-21
Create nodejs.yml

trcywu:


frigus02:
I had an early look. I think this is great. Choice of libraries looks perfect. A couple of smaller points I noticed:
- We want TypeScript `strict: true`
- We don't need to test with multiple node versions. The latest (maybe LTS?) is enough
- Tests can probably run on Ubuntu instead of Mac. Our cluster will run on Ubuntu, so running on Ubuntu is closer to that environment

frontend/tsconfig.json
```diff
@@ -0,0 +1,19 @@
+{
+  "compilerOptions": {
+    "target": "es5",
+    "lib": ["dom", "dom.iterable", "esnext"],
+    "allowJs": true,
+    "skipLibCheck": true,
+    "strict": false,
```
frigus02, line 7: I think we definitely want `"strict": true`

frontend/package-lock.json
```diff
@@ -0,0 +1,10870 @@
+{
```
frigus02, line 1: We're using `yarn`, so we don't need this file.

PR #4, opened 2020-05-28
Infrastructure setup

matt-paul:
- Setup Terraform configuration for non-production environment
- Includes:
  - Virtual Network
  - Azure Kubernetes Service
- Tooling to enable every developer to have their own instance of the environment in non-production to enable independent testing

TODO: [NOTE: We are doing these in further PR's]
- [ ] Create README
- [ ] Add Container registry
- [ ] Validate Terraform in CI


infrastructure/scripts/create-storage-account.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
+
```
matt-paul, line 2: Perhaps we should rename this script 'set up dev cluster' or something, as the current name doesn't cover everything?

infrastructure/scripts/create-storage-account.sh
```diff
@@ -0,0 +1,36 @@
...
+set -eu
+
+NAME="${1:?"Please enter your name as first argument"}"
+
+RESOURCE_GROUP_NAME=tfstate$NAME
```
matt-paul, line 7: Does this make sense as a name? 🤔

alsuren: seems fine: it puts them all in a similar namespace, and includes your name, so offboarding devs sounds quite easy (just go to https://portal.azure.com/#blade/HubsExtension/BrowseResourceGroups and search for their name)

I guess we could have a - in the middle of it maybe?

matt-paul: I'm just unclear on the relevance of 'tfstate' to a resource group....

frigus02: We talked about this very briefly. It makes sense to have the Terraform in a different resource group. I can think of instances were we want to delete all of the infrastructure (which is easily done by deleting the resource group), but without deleting the Terraform state.

Resource groups don't cost anything so there is not really a downside.

I agree that we could put a `-` in the middle. We seem to do this for other resources, too.

frigus02: For reference: We found out that at least for the storage account we can't use hyphens in the name. It has to consist of 3 - 24 lowercase letters or numbers.

infrastructure/scripts/create-dev-environment.sh
```diff
@@ -0,0 +1,34 @@
...
+az storage container create --name $CONTAINER_NAME --account-name $STORAGE_ACCOUNT_NAME --account-key "$ACCESS_KEY"
+
+echo "
+# Create a terraform.tfvars file (in infrastructure/environments/dev/) with the following content:
+
+resource_group_name=\"$RESOURCE_GROUP_NAME\"
+storage_account_name=\"$STORAGE_ACCOUNT_NAME\"
+USERNAME=\"$NAME\"
+
+# and then run `terraform init -backend-config=terraform.tfvars` in the same folder.
```
frigus02, line 33: This actually runs terraform because of the backticks. Maybe use this instead:
```suggestion
# and then run \"terraform init -backend-config=terraform.tfvars\" in the same folder.
```

PR #5, opened 2020-06-01
Create README for Terraform Setup

cobwebfx:


infrastructure/README.md
```diff
@@ -0,0 +1,69 @@
...
+  ```bash
+terraform plan
+  ``` 
+
+Apply changes and execute your plan on the server
+  ```bash
+tf apply
+  ``` 
+
+Remember to teardown your environment when you're done
```
matt-paul, line 66: Maybe just add the reason why destroying the environment is important 🙂

infrastructure/README.md
```diff
@@ -0,0 +1,69 @@
...
+./infrastructure/scripts/create-dev-environment.sh john
+``` 
+  
+CD into your new dev folder
+```bash
+./infrastructure/environments/dev
+``` 
+  
+Follow the onscreen instructions to create the terraform.tfvars file.
+The contents of the file should still be printed on the screen.
```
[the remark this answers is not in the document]
cobwebfx: It's not a screenshot, but I've pasted how it should show up.  See code block a couple of lines below this..

infrastructure/README.md
```diff
@@ -0,0 +1,69 @@
...
+  
+  Make sure you have the correct version of Terraform.  
+  For this project, it is 0.12.25
+  ```bash
+terraform --version
+  ```  
+
+Adjust it if nessassary.  Note, you may need CHTF installed.
+  ```bash
+chtf 0.12.25
```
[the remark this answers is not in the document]
cobwebfx: I've replaced CHTF with Terraform Switcher as the default, but left CHTF as an option, due to the fact that it is widely used, and gives the user a second option if they struggle with the installation.

infrastructure/README.md
```diff
@@ -0,0 +1,76 @@
...
+terraform plan
+  ``` 
+
+Apply changes and execute your plan on the server
+  ```bash
+terraform apply
+  ``` 
+
+You will be billed for the duration of this environment being up.
+So you must teardown your environment when you're done.
```
frigus02, line 73: That almost sounds like we bill the developer. Maybe we can reword this to sound less scary. It's really more about being a bit cautious with costs. E.g.

"To reduce infrastructure costs for the NHS, please destroy your environment when you no longer need it."

infrastructure/README.md
```diff
@@ -0,0 +1,104 @@
...
+   ```    
+1. Install [Azure CLI](https://docs.microsoft.com/en-us/cli/azure/install-azure-cli?view=azure-cli-latest) 
+    ```bash
+    brew install azure-cli
+    ```      
+1. Login to Azure 
+      ```bash
+    az login
+      ``` 
+1. Install Git
```
frigus02, line 41: I don't think we need to include that here. You need git if you want to do any contributions to the repository.

infrastructure/README.md
```diff
@@ -0,0 +1,104 @@
...
+      ```bash
+    brew install git
+      ``` 
+  
+## Dev Environment Setup
+ 
+Infrastructure is setup so every developer can create their own instance of the environment in Azure,
+as opposed to sharing a staging environment.
+
+1. Clone the FutureNHS Platform
```
frigus02, line 51: I don't think we need that and the point about `cd platform`, either. I would prefer to keep the instructions specific to the infrastructure. If it turns out that we need basic instructions of how to get started with the repository, we could create a separate file link to it.

infrastructure/README.md
```diff
@@ -0,0 +1,111 @@
...
+    cd infrastructure/environments/dev
+    ``` 
+  
+1. Follow the onscreen instructions to create the terraform.tfvars file
+
+    The contents of the file should still be printed on the screen
+
+    Below is an example of what it would likely look like:
+    
+    ```bash
```
frigus02, line 82: That's not actually `bash` in this case but `terraform`.

infrastructure/README.md
frigus02: I think we don't need the instruction to clone, either.

cobwebfx: Are you sure?  I've found the clone command to be a standard addition in README files.  It's also the first step in the setup.  If you still want me to delete it after this, I will.  I just wanted to give my opinion.

frigus02: Hm. We can put it in the prerequiresites at the top. At some point we might also want to describe how to apply changes to production, which also requires you to clone the repository.

I just think cloning the repository is not something special for infrastructure. You will need to do that to run almost anything. I would have added this to general contribution guidelines, if at all. 🤔

frigus02: I thought about it a bit more. Maybe it's not a bad idea to have the step here. Maybe we just remove the `git clone` command. Make it less noisy and one less URL that we have to keep up to date.

matt-paul:
One more thing that i have realised we are missing here - in order to interact with the cluster (ie kubectl commands) that has been created, we need to run the following after the terraform apply.
`az aks get-credentials --resource-group platform-dev-$DEVELOPER_NAME --name dev-$DEVELOPER_NAME`
`

cobwebfx:
@matt-paul can you run me through this when you get in?

PR #9, opened 2020-06-03
Build Docker image

trcywu:


frontend/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM node:12
+
+# Setting working directory. All the path will be relative to WORKDIR
+WORKDIR /usr/src/app
+
+# Installing dependencies
+COPY package*.json ./
+RUN npm install
```
frigus02, line 8: We're using `yarn` and we have `yarn.lock` file. Let's use that here, too.
```suggestion
COPY package.json yarn.lock ./
RUN yarn install --frozen-lockfile
```

Same further down in the file.

.github/workflows/nodejs.yml
```diff
@@ -31,3 +30,23 @@ jobs:
           yarn lint
           yarn test
           yarn build
```
frigus02, line 32: Let's move these steps into the Docker build, too. This way we only run them once.

This means we don't even need to setup node on the build agent, because everything runs in Docker.

frontend/manifests/deployment.yml
```diff
@@ -0,0 +1,40 @@
...
+          image: citracytest.azurecr.io/frontend
+          ports:
+            - containerPort: 3000
+          resources:
+            limits:
+              cpu: 100m
+              memory: 100Mi
+            requests:
+              cpu: 10m
+              memory: 100Mi
```
frigus02, line 26: We might want to remove those, but we can do it later. I imagine the node app could require more resources than this.

infrastructure/README.md
```diff
@@ -6,127 +6,129 @@ We use [Terraform](https://www.terraform.io/) to build our environments.
...
 1. To reduce infrastructure costs for the NHS, please destroy your environment when you no longer need it.
 
-    ```bash
-    terraform destroy
-    ``` 
+   ```bash
+   terraform destroy
+   ```
+
+   At the same time, delete ./infrastructure/environments/dev/.terraform folder as it caches your existing terraform setup and if left untouched will create an error next time you try to spin up a new cluster.
```
frigus02, line 134: That does not happen on my machine. This should not be necessary. Deleting the `.terraform` folder has the downside that you need to run `terraform init` again next time.

Can you try this again and see if this is really necessary?

trcywu: Matt and I both got the same error when we were trying to spin up our clusters. When we went through and debugged it was because there was a cached version in .terraform so deleting it resolved the error. Can delete for now and if the error comes again at least we know what it is.

frigus02: Was your cached version still from our mobbing session? We changed the tfstate storage account after that. So that would be an explanation.

trcywu: ah yes, it was from the mobbing session.

frigus02: I'm pretty sure that was the reason. As long as we don't change the tfstate config anymore (which we won't 🤞), this should not be necessary.

Or: This step is only necessary if the backend config changed.

Let's remove this as an explicit step here. It could be a good idea to add a `## Troubleshooting` section at the end of this file, though. We can add this tip there.

infrastructure/environments/production/main.tf
```diff
@@ -23,3 +23,10 @@ module platform {
   environment = "production"
   location    = var.location
 }
+
+resource "azurerm_container_registry" "acr" {
+  name                = "fnhsproduction"
+  resource_group_name = "platform-production"
```
frigus02, line 29: I think we need to add a dependency here. The resource group is created inside the platform module. But terraform is not smart enough to figure this out on it's own.

I suggest we add an `output` variable in the platform module, e.g.

```ts
output "resource_group_name" {
  value = azurerm_resource_group.platform.name
}
```

Then we can use it here:

```ts
   resource_group_name = module.platform.resource_group_name
```

infrastructure/README.md
```diff
@@ -6,127 +6,131 @@ We use [Terraform](https://www.terraform.io/) to build our environments.
...
 1. Install Terraform:
 
-    ```bash
-    brew install terraform
-    ```  
-   
+   ```bash
+   brew install terraform
+   ```
+
```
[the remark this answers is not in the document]
matt-paul: Hmm @trcywu @alsuren , sorry i'm late too this, i did not recieve an email notification, and have so only seen this in the notifications dashboard i'm starting to play with.  Firstly, is this still unresolved, and secondly i find the question confusing ( unsure relevence of IDE plugin ?) - could you rephrase / elaborate please if you still need input?

PR #12, opened 2020-06-05
Hello world

matt-paul:
The aim of this PR is to get a very simple app into our cluster.  This will allow us to move onto other tickets such as setting up our tracing solution.

- [x] Shamelessly lift app from [MHRA](https://github.com/MHRA/products).
- [x]  Create Github workflow to build and push image to Azure Container Registry.
- [x] Create temporay k8s deployment (deployments will live in separate repo at a later date but can live with the app for now.)


.github/workflows/hello-world.yml
```diff
@@ -0,0 +1,43 @@
...
+        with:
+          path: platform
+
+      - name: "Login via Azure CLI"
+        uses: azure/login@v1
+        with:
+          creds: ${{ secrets.AZURE_CREDENTIALS }}
+      - name: "Login to docker"
+        uses: azure/docker-login@v1
+        with:
```
matt-paul, line 31: Standing on the shoulders of giants here, thanks Tracy!

.github/workflows/hello-world.yml
```diff
@@ -0,0 +1,39 @@
...
+on:
+  push:
+    branches:
+      - "**"
+    paths:
+      - hello-world/**
+      - .github/workflows/hello-world.yml
+
+env:
+  IMAGE: citracytest.azurecr.io/platform/hello-world
```
matt-paul, line 12: As mentioned, we can either merge this in and change the container registry later once set up, or hold off merging this PR

hello-world/Makefile
```diff
@@ -0,0 +1,32 @@
...
+
+.PHONY: docker-build
+docker-build: ## Build and tag Docker image
+	DOCKER_BUILDKIT=1 docker build \
+		--progress plain \
+		--tag $(app) \
+		--tag $(image) \
+		--tag $(image):$(tag) \
+		.
+.PHONY: docker-run
```
frigus02, line 21: There should probably be a empty line before this

hello-world/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "hello-world"
+version = "0.1.0"
+authors = ["stuart.harris@red-badger.com"]
```
alsuren, line 4: This entire folder smells like copy-pasta. Can you add a README explaining where you got it from, and/or attribution comments at the top of each file? You may also need to copy a LICENSE file into this directory.

.gitignore
```diff
@@ -1,3 +1,5 @@
 .env
 .terraform/
 terraform.tfvars
+
+target
```
frigus02, line 5: That's a folder, isn't it? Should we add a slash to make this easier to understand? `target/`

hello-world/Cargo.toml
matt-paul: Hmm - i deliberately left Stu here because of that.  That's not enough attribution? Good shout about the README though

matt-paul: ah, see it's 'and / or' 👍

hello-world/Makefile
```diff
@@ -0,0 +1,32 @@
...
+	docker run -it -p 3030:3030 $(image)
+
+.PHONY: docker-push
+docker-push: ## Push Docker image
+	docker push $(image)
+	docker push $(image):$(tag)
+
+.PHONY: help
+help: ## Display this help screen
+	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' $(MAKEFILE_LIST) | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-30s\033[0m %s\n", $$1, $$2}'
```
alsuren, line 32: maybe chop this line up with \ like how line 15 is done? (also add attribution? Feels quite similar to https://gist.github.com/prwhite/8168133#gistcomment-3291344 )

hello-world/Cargo.toml
matt-paul: As discussed, will add the MIT License from MHRA into this directory and add a note in a README.

PR #13, opened 2020-06-05
Add Test Approach doc

declanslevin:
Adding the [FutureNHS Test Approach](https://docs.google.com/document/d/1n82RsYtKDAxh21GfbI_0Ocy0YV16249IdxbP_g0knS8/edit?usp=sharing) doc to the repository with some modifications and a table of contents. 

Homer Simpson burning a document


docs/testing/test_approach.md
```diff
@@ -0,0 +1,305 @@
...
+
+Further testing information:
+https://accessibility.18f.gov/checklist/
+
+We might need to get an accessibility audit according to this https://www.gov.uk/service-manual/helping-people-to-use-your-service/testing-for-accessibility#getting-an-accessibility-audit
+
+## Issue Logging and Handling
+Anyone who discovers an issue should feel free to submit it on GitHub. This way the issues sit alongside the codebase and can be resolved automatically upon a relevant bug-fix being merged. When an issue is logged it should be added to the project backlog for prioritisation. If the person logging an issue feels it has enough urgency they should raise it with the product owner/wider team for discussion. Otherwise, regular ‘triage’ sessions will be held to discuss the bug backlog and whether there are any issues that can be brought into the current stream of work to be fixed.
+
+#### Triage
```
frigus02, line 271: I think this should be a heading level 3
```suggestion
### Triage
```

Similar to the following headings (Feature Requests, Tech Debt, Writing a Unit Test) and the ones under Testing Tooling (Automation, Manual Testing aids)

declanslevin:
> Looks great.
> 
> Matt made a good point. Some of the decisions in here would be worth calling out specifically in an architecture decision record. That way they are easier to parse/spot. I can see 2 things here that make sense in my mind:
> 
> * Feature flagging
> * No staging environment / only production
> 
> Should we add those in a follow-up PR? Then we can link to them from this testing approach document.

I can do that, would this go under /architecture/decisions?

PR #14, opened 2020-06-09
Adding feature flags and no staging to architectural decision record

declanslevin:


frigus02:
Can you use the same format as the other decision records? We're going with this here: [adr.github.io](https://adr.github.io) (see also [0001-record-architecture-decisions.md](https://github.com/FutureNHS/futurenhs-platform/blob/master/docs/architecture/decisions/0001-record-architecture-decisions.md)).

We can possibly split this into 2 different decisions:
1. We're going to use feature flagging
2. We're going to have only 1 long lived environment: production (made possible by decision about feature flagging)

PR #16, opened 2020-06-11
Tracing: Hello world app with `tracing`

matt-paul:
## Set up tracing within demo app.

[Link to airtable ticket](https://airtable.com/tblgtzV3sSwquTkIa/viw8ZHoNqvSvl62lt/recEBTYlei5CmMkR7?blocks=hide)

## Acceptance Criteria:
   - [x] Hello world app is instrumented with [tracing](https://github.com/tokio-rs/tracing) crate. A good overview of the create can be found [here](https://tokio.rs/blog/2019-08-tracing/). This is combined with [opentelemetry-rust](https://github.com/open-telemetry/opentelemetry-rust), and [application insights open telemetry exporter](https://github.com/frigus02/opentelemetry-application-insights).
   - [x] K8s [ConfigMap](hello-world/dev/container-azm-ms-agentconfig.yaml) for collecting logs is applied.
   - [x] Pod logs can be seen in [Azure Monitor](https://portal.azure.com/#blade/Microsoft_Azure_Monitoring/AzureMonitoringBrowseBlade/overview).
 - [x] Add application insights app to Terraform.
 - [x] Use [tokio](https://github.com/tokio-rs/tokio) in place of [smol](https://docs.rs/smol/0.1.18/smol/) as async runtime to allow for batch span processing.

## Testing
1) Within `infrastructure/environments/dev` directory, run `terraform apply` to create your dev cluster. Make a note of the 'instrumentation key' that is output.
2) Within `hello-world/dev` directory, add the instrumentation key from above to the deployment environmental variable `INSTRUMENTATION_KEY` and then run `kubectl apply -f deployment.yaml`.
1) `kubectl get pods` - get the name of the hello world pod in your cluster.
2) `kubectl port-forward <my-pod-name> 3030:3030` - port forwarding for localhost testing.
3) Proceed to localhost:3030/hello/ and append something.
4) Go to Azure monitor and verify you see traces including the 'something' from step 5.

Screenshot 2020-06-23 at 10 36 13


hello-world/Cargo.toml
```diff
@@ -7,8 +7,15 @@ edition = "2018"
...
 futures = "0.3.5"
-smol = "0.1.4"
-tide = "0.8.1"
+opentelemetry-application-insights = "0.1.0"
+opentelemetry = "0.6"
+tide = "0.11.0"
+tokio = { version = "0.2", features = ["macros", "time", "rt-core", "stream"]}
+tracing = { version = "0.1", features = ["attributes"] }
+tracing-futures = "0.2.4"
+tracing-log = "0.1.1"
```
frigus02, line 16: Are we using `tracing-log` anywhere? We probably don't need that.

hello-world/dev/container-azm-ms-agentconfig.yaml
```diff
@@ -0,0 +1,44 @@
+kind: ConfigMap
+apiVersion: v1
+data:
+  schema-version:
+    #string.used by agent to parse config. supported versions are {v1}. Configs with other schema versions will be rejected by the agent.
+    v1
+  config-version:
+    #string.used by customer to keep track of this config file's version in their source control/repository (max allowed 10 chars, other chars will be truncated)
+    ver1
+  log-data-collection-settings: |-
```
frigus02, line 10: We're passing uploading traces directly to application insights now. What's the reason for having this log collection? I could possibly see this being useful for other services like Argo CD or NGINX, which don't integrate with Application Insights directly. In this case, maybe we want to exclude the "default" namespace.

The file should also live somewhere central. It's not specific to hello world. This config map can only exist in the cluster once. Let's move it to `infrastructure/kubernetes/logging/` or something similar. And let's also document that it should be applied to the cluster.

hello-world/dev/deployment.yaml
```diff
@@ -38,3 +38,6 @@ spec:
             initialDelaySeconds: 5
             periodSeconds: 20
             timeoutSeconds: 30
+          env:
+          - name: INSTRUMENTATION_KEY
+            value: ""
```
frigus02, line 43: Should we add a comment for the instrumentation key describing where to get it from? E.g.
> Application Insights instrumentation key. Printed after applying dev or production terraform environment.

hello-world/dev/container-azm-ms-agentconfig.yaml
```diff
@@ -0,0 +1,44 @@
...
+    # Log data collection settings
+    # Any errors related to config map settings can be found in the KubeMonAgentEvents table in the Log Analytics workspace that the cluster is sending data to.
+    [log_collection_settings]
```
alsuren, line 13: Oh my god. It's toml nested within yaml.

hello-world/src/lib.rs
```diff
@@ -1,16 +1,48 @@
...
 
-pub fn create_app() -> Server<()> {
-    let mut app = tide::new();
     app.at("/hello/:name").get(hello_handler);
     app
 }
 
-async fn hello_handler(req: Request<()>) -> tide::Result<String> {
+#[tracing::instrument]
+async fn hello_handler(req: Request<State>) -> tide::Result<String> {
```
alsuren, line 38: out of curiosity, how many of `#[tracing::instrument]` / `.instrument(span.clone())` can we leave out before the azure ui stops being able to link up the `info!("The users name is: {}", name)` and the request's `http.status_code`?

hello-world/dev/container-azm-ms-agentconfig.yaml
alsuren: ... (optional) These all look like default settings. I would be tempted to just link to the docs and only include the values that are not set to their default values.

frigus02: Yeah, I thought about that as well. It might actually be easier to read if we removed all the comments and default settings.

hello-world/src/lib.rs
frigus02: We only need then`.instrument(span.clone())` in the middleware. All other `#[tracing::instrument]` just add intermediary spans and are entirely optional.

hello-world/dev/container-azm-ms-agentconfig.yaml
matt-paul: Makes sense, will do

hello-world/src/lib.rs
frigus02: I think we should do a lunch and learn or a mobbing session or something to get the entire team across how this tracing stuff works.

We are using OpenTelemetry. The best documentation is probably the [trace API spec](https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/trace/api.md). In this hello world example it's wrapped in the `tracing` crate, which provides some nice macros for collecting spans. For Next.js we can use [opentelemetry-js](https://github.com/open-telemetry/opentelemetry-js), which automatically instruments HTTP calls and I think also express.js.

infrastructure/README.md
```diff
@@ -143,6 +143,10 @@ as opposed to sharing a staging environment.
 
    and browse to http://localhost:3000.
 
+1.  Apply the ConfigMap for Azure Monitor for Containers to collect data in the Log Analytics workspace.  The ConfigMap can be found in `infrastructure/kubernetes/logging` directory.
+   ```bash
+   kubectl create configmap --from-file= container-azm-ms-agentconfig.yaml
```
frigus02, line 148: Let's use `kubectl apply` because that also works for updates.

```suggestion
   kubectl apply -f container-azm-ms-agentconfig.yaml
```

infrastructure/kubernetes/logging/container-azm-ms-agentconfig.yaml
```diff
@@ -0,0 +1,24 @@
+kind: ConfigMap
```
frigus02, line 1: Should we add a comment on top, linking to the documentation?

https://docs.microsoft.com/en-us/azure/azure-monitor/insights/container-insights-agent-config#configmap-file-settings-overview

infrastructure/README.md
matt-paul: Ah, i had not realised that, thanks

PR #19, opened 2020-06-16
Fix build trigger for path filters

frigus02:
When using the path filters, GitHub generates a list of changed files differently for pull request and push triggers (see [docs](https://help.github.com/en/actions/reference/workflow-syntax-for-github-actions#git-diff-comparisons)):

- Pull request: diff between most recent commit and merge base
- Push: diff between head and base commit

This means our builds were not triggered if the most recent push didn't contain changes to the corresponding files, even if the overall pull request did include changes to them.

This changes the logic to use a pull request based trigger. Pushes itself won't trigger builds now. If this turns out to be an issue, we can enable both.

### DIFF
--- a/.github/workflows/frontend.yml
+++ b/.github/workflows/frontend.yml
@@ -1,6 +1,3 @@
-# This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
-# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
-
 name: frontend
 
 on:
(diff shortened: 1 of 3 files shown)


alsuren:
> Pushes itself won't trigger builds now.

by this, do you mean 
a) pushing to a branch won't trigger a build until you make a draft pull request
b) pushing more commits to a branch (even after you've made a pull request) won't trigger a re-build?
c) both

I'm assuming it's a) so :ship: once you fix the build failure.

frigus02:
Ah yes. I mean only a). It only builds if you make a PR. But then it will build on every push.

PR #26, opened 2020-06-22
Setup load balancer

cobwebfx:
- [x] Hello World Service Kubernetes Manifest
- [x] Ingress Kubernetes Manifests
- [x] TLS Certificate Kubernetes Manifest
- [x] Kustomize Kubernetes Manifests
- [x] Kustomize DNS Patch Kubernetes Manifest
- [x] Setup Ingress Dev Environment Script
- [x] Teardown Ingress Dev Environment Script


infrastructure/environments/dev/k8s/.gitignore
```diff
@@ -0,0 +1 @@
+.temp
```
alsuren, line 1: let's get into the habit of adding newlines to the ends of our files. You should be able to set up your editor to do this.

infrastructure/environments/dev/setupTf.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
```
alsuren: Maybe merge this into scripts/create-dev-environment.sh ? I would also be tempted to delete the .terraform directory if we're going nuclear with this.

infrastructure/environments/dev/k8s/.gitignore
cobwebfx: Can do.  Is that a biggie for this PR, as this will be the case for a lot of files?

infrastructure/scripts/setup-dev-ingress.sh
```diff
@@ -0,0 +1,48 @@
...
+sed -i "" "s/--DNS_PREFIX--/$NAME/g" "$CURRENT_DIR/../environments/dev/k8s/.temp/ingress.yaml"
+sed -i "" "s/--REGION--/$REGION/g" "$CURRENT_DIR/../environments/dev/k8s/.temp/ingress.yaml"
+sed -i "" "s/--EMAIL--/$EMAIL/g" "$CURRENT_DIR/../environments/dev/k8s/.temp/ingress.yaml"
+sed -i "" "s/--USERNAME--/$NAME/g" "$CURRENT_DIR/../environments/dev/k8s/.temp/certificate.yaml"
+sed -i "" "s/--DNS_PREFIX--/$NAME/g" "$CURRENT_DIR/../environments/dev/k8s/.temp/certificate.yaml"
+sed -i "" "s/--REGION--/$REGION/g" "$CURRENT_DIR/../environments/dev/k8s/.temp/certificate.yaml"
+sed -i "" "s/--EMAIL--/$EMAIL/g" "$CURRENT_DIR/../environments/dev/k8s/.temp/certificate.yaml"
+
+echo " > Setup Hello World Service"
+kubectl apply -f "$CURRENT_DIR/../../hello-world/dev"
```
alsuren, line 28: do you mean `hello-world/dev/cluster-ip-service.yaml` or something?

infrastructure/scripts/setup-dev-ingress.sh
```diff
@@ -0,0 +1,48 @@
...
+kubectl apply -f "$CURRENT_DIR/../environments/dev/k8s/.temp/ingress.yaml"
+
+echo " > Create cert-manager namespace"
+kubectl create ns cert-manager
+
+echo " > Add Jetstack repos to Helm"
+helm repo add jetstack https://charts.jetstack.io
+
+echo " > Install cert-manager from Helm"
+helm install cert-manager --namespace cert-manager jetstack/cert-manager
```
alsuren, line 40: as long as @frigus02 is okay with helm

infrastructure/environments/dev/k8s/ingress/ingress-install.yaml
```diff
@@ -0,0 +1,647 @@
+apiVersion: v1
```
alsuren, line 1: This feels generated, or copy-pasted from somewhere. Might be nice to have some docs about where it comes from.

infrastructure/environments/dev/k8s/.gitignore
alsuren: Just fix the ones that you added before you merge, to reduce conflicts down the line. We can work on the rest later.

infrastructure/environments/dev/k8s/ingress/ingress-install.yaml
frigus02: I just saw that this installs NGINX with only 1 replica. We may have to use helm instead to easily increase the replica count. I personally would use helm only to generate this file. But I'm happy to hear what about think about it.

```sh
helm repo add stable https://kubernetes-charts.storage.googleapis.com/

helm template nginx stable/nginx-ingress \
    --set controller.replicaCount=2 \
    --set controller.nodeSelector."beta\.kubernetes\.io/os"=linux \
    --set defaultBackend.nodeSelector."beta\.kubernetes\.io/os"=linux
```

infrastructure/scripts/setup-dev-ingress.sh
frigus02: Yeah, I'm not sure. I think I would use helm to generate YAML files and then commit those files. We can do that with:

```sh
helm repo add jetstack https://charts.jetstack.io

helm template cert-manager jetstack/cert-manager
```

I think we may also have to include the custom resource definitions. We could do that with:

```sh
helm template cert-manager jetstack/cert-manager \
    --set installCRDs=true
```

infrastructure/scripts/setup-dev-ingress.sh
frigus02: You can point `kubectl` to a directly and it will apply all files inside it. I think it does not recurse, though. So this applies the service and the deployment.yml.

infrastructure/scripts/setup-dev-ingress.sh
```diff
@@ -0,0 +1,48 @@
...
+sed -i "" "s/--EMAIL--/$EMAIL/g" "$CURRENT_DIR/../environments/dev/k8s/.temp/certificate.yaml"
+
+echo " > Setup Hello World Service"
+kubectl apply -f "$CURRENT_DIR/../../hello-world/dev"
+
+echo " > Setup Ingress Service"
+kubectl apply -f "$CURRENT_DIR/../environments/dev/k8s/.temp/ingress.yaml"
+
+echo " > Create cert-manager namespace"
+kubectl create ns cert-manager
```
frigus02, line 34: I would prefer to use `kubectl apply` for the namespace, too. We can copy the approach from what Tracy did for Argo CD: https://github.com/FutureNHS/futurenhs-platform/blob/master/infrastructure/kubernetes/argocd-setup/namespace.yaml

infrastructure/environments/dev/setupTf.sh
cobwebfx: I accidentally pushed this.  It's just a script I have locally to save time.

infrastructure/scripts/setup-dev-ingress.sh
cobwebfx: I'm executing all kubernetes manifests in the directory.  So, yeah, the cluster-ip-service.yaml and deployment.yaml.

infrastructure/README.md
```diff
@@ -143,8 +158,21 @@ as opposed to sharing a staging environment.
...
+    and registers a TLS Certificate with your temporary domain. 
+    
+    ```bash
+    ./infrastructure/scripts/setup-dev-ingress.sh $FNHSNAME $FNHSEMAIL
+    ```
+
 1. To reduce infrastructure costs for the NHS, please destroy your environment when you no longer need it.
 
+   ```bash
+   ./infrastructure/scripts/teardown-dev-ingress.sh
```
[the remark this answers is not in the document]
cobwebfx: I can remove.  No, Terraform doesn't destroy the load balancer.  I found that running the commands in this bash script prevents the issues we were having with the Public IPs failing.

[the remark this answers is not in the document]
frigus02: Oh no, that's really annoying. To me it looked like the problems we saw were due to all IPs using the same DNS name, which clashed. I might test this as well, just to make sure. It would be good to understand if the script is really necessary.

infrastructure/scripts/setup-dev-ingress.sh
```diff
@@ -0,0 +1,43 @@
...
+echo " > Set vars in build file"
+sed -i "" "s/--USERNAME--/$NAME/g" "$CURRENT_DIR/../environments/dev/k8s/.temp/ingress.yaml"
+sed -i "" "s/--DNS_PREFIX--/$NAME/g" "$CURRENT_DIR/../environments/dev/k8s/.temp/ingress.yaml"
+sed -i "" "s/--REGION--/$REGION/g" "$CURRENT_DIR/../environments/dev/k8s/.temp/ingress.yaml"
+sed -i "" "s/--EMAIL--/$EMAIL/g" "$CURRENT_DIR/../environments/dev/k8s/.temp/ingress.yaml"
+sed -i "" "s/--USERNAME--/$NAME/g" "$CURRENT_DIR/../environments/dev/k8s/.temp/certificate.yaml"
+sed -i "" "s/--DNS_PREFIX--/$NAME/g" "$CURRENT_DIR/../environments/dev/k8s/.temp/certificate.yaml"
+sed -i "" "s/--REGION--/$REGION/g" "$CURRENT_DIR/../environments/dev/k8s/.temp/certificate.yaml"
+sed -i "" "s/--EMAIL--/$EMAIL/g" "$CURRENT_DIR/../environments/dev/k8s/.temp/certificate.yaml"
+
```
cobwebfx, line 27: @frigus02 the issue with the certificate-issuer manifest is that it is never executed.  
When the ingress manifest is executed, it looks for the Issuer in the same directory.
So the ingress.yaml and the certificate.yaml need to be in the same directory.
If you try to execute the certificate-issuer manifest, kubectl errors with "no matches for kind Issuer"
Unless you have a way that I'd not thought of?  I can explain in a call if this is not clear.
(note. When I say "executed", I mean "kubectl apply"

infrastructure/environments/dev/k8s/ingress/ingress-install.yaml
cobwebfx: @frigus02 how do you feel about this solution instead:
infrastructure/environments/dev/k8s/ingress/patch-replica.yaml

infrastructure/README.md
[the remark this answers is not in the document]
frigus02: I just tested this. I created my cluster, then I installed the cert manager and nginx ingress controller from this branch. i waited until everything was up and running.

Then I just ran `terraform destroy`. It completed successfully after 12mins. Deleted all IPs without complaining.

My hopes are up again, that we don't need that script 🙂 🤞

infrastructure/kubernetes/ingress/step2/prod/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - ../base/
+patchesStrategicMerge:
+  - cert-issuer-patch-env.yaml
```
frigus02, line 6: Should we patch the ingress here, to get rid of the `--DNS_PREFIX--` and hardcode `fnhs`?

infrastructure/README.md
```diff
@@ -53,6 +53,18 @@ We use [Terraform](https://www.terraform.io/) to build our environments.
    ```bash
    brew install linkerd
    ```
+   
+1. Install [Helm](https://helm.sh/docs/intro/install/):
```
frigus02, line 57: No longer needed 🎉

infrastructure/scripts/setup-ingress.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+
+set -eu
+
+NAME="${1:?"Please enter your name as first argument"}"
+EMAIL="${2:?"Please enter your email address as second argument"}"
+ENV="${3:-dev}"
+REGION="${4:-westeurope}"
+CURRENT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" >/dev/null 2>&1 && pwd)"
+DNS_PREFIX=$NAME
```
frigus02, line 10: I wonder if the DNS prefix should be `fnhs-dev-$NAME` for dev environments. I think the prefix has to be globally unique. So if anyone already registered `jan`, then it would fail for me. With an `fnhs` prefix, this is less likely.

infrastructure/scripts/setup-ingress.sh
```diff
@@ -0,0 +1,49 @@
...
+if [ $ENV == "prod" ]; then
+  DNS_PREFIX="fnhs"
+fi
+
+echo " > Setup Hello World Service"
+# hello-world/dev used because currently no prod version
+kubectl apply -f "$CURRENT_DIR/../../hello-world/dev"
```
frigus02, line 17: The hello world service is going to be deployed by Argo CD. Let's remove that from here.

I hope that's possible and Kubernetes doesn't complain when applying the Ingress.

infrastructure/environments/dev/k8s/.gitignore
alsuren: looks like we don't have a .temp dir anymore, so this can be deleted

infrastructure/scripts/setup-ingress.sh
alsuren: I would be tempted to get all of the currently in-flight stuff merged as-is, and then have a mobbing session where we go around adding TODO items to move things about.

infrastructure/README.md
```diff
@@ -95,24 +114,20 @@ as opposed to sharing a staging environment.
...
+   
+    ```bash
+    az aks get-credentials --resource-group platform-dev-$FNHSNAME --name dev-$FNHSNAME
+    ``` 
 
 1. To install the [Linkerd](https://linkerd.io/) control plane, run the `install-linkerd.sh` script that can be found within `infrastructure/scripts` directory.
 
    ```bash
-   ./install-linkerd.sh
+   ./infrastructure/scripts/install-linkerd.sh
```
alsuren, line 130: 🍰  I'm glad I'm not the only one that likes running everything from the root of the git repo.

infrastructure/scripts/setup-ingress.sh
frigus02: Fine with me. 🙂

infrastructure/README.md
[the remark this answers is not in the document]
frigus02: This is no longer needed.

infrastructure/README.md
```diff
@@ -146,9 +155,23 @@ as opposed to sharing a staging environment.
...
    kubectl apply -f container-azm-ms-agentconfig.yaml
+   ```
+
+1. Spin up an Ingress Controller with your **name** and **email** as parameters. 
+
+    This also creates a Load Balancer with a Public IP, creates a temporary domain, 
+    and registers a TLS Certificate with your temporary domain. 
+    
+    ```bash
+    ./infrastructure/scripts/setup-dev-ingress.sh $FNHSNAME $FNHSEMAIL
```
frigus02, line 166: That's now called `setup-ingress.sh`.

PR #27, opened 2020-06-23
Connecting Deployment repo to ArgoCD

trcywu:
## Name of story / Ticket #
Link to Zeplin/design files

## Acceptance Criteria

1. Clone platform repo
2. Docker login
3. Build & push docker image, with tag using commit sha 
4. Clone deployments repo
5. Copy over frontend/manifest deployment.yml into the argocd(deployments) repo
6. Kustomize the new copied yaml and find new image by commit sha from step 3
7. Deploy to Argo CD - find the above tag 

## Pre-review checklist:

- [ ] Tests

- [ ] Documentation

- [ ] Feature flag

## Testing information

- Test plan?

## Test:

- [ ] Tested locally 

- Platforms/Browsers: 

  - [ ] Internet Explorer 11
  
  - [ ] Edge
  
- Accessibility:

  - [ ] Screen Reader
  
  - [ ] Keyboard Navigation
  
  - [ ] Magnification
  
  - [ ] Contrast
  
  - [ ] Text size 200%
  
  - [ ] Landscape (Mobile)


.github/workflows/frontend-master.yml
```diff
@@ -0,0 +1,59 @@
...
+        with:
+          repository: FutureNHS/futurenhs-deployments
+          path: futurenhs-deployments
+          #github deploy key?
+      - name: Clone Deployment yaml
+        uses: actions/checkout@v2
+        with:
+          repository: FutureNHS/futurenhs-platform
+          path: futurenhs-platform/frontend/manifests/deployement.yml
+          #do something here?
```
alsuren, line 39: I don't think you need this second one

.github/workflows/frontend-master.yml
```diff
@@ -0,0 +1,59 @@
...
+        run: |
+          cd frontend
+          docker build . -t ${{ secrets.REGISTRY_LOGIN_SERVER }}/frontend:${{ github.sha }}
+          docker push ${{ secrets.REGISTRY_LOGIN_SERVER }}/frontend:${{ github.sha }}
+      - name: Clone Deployments repo
+        uses: actions/checkout@v2
+        with:
+          repository: FutureNHS/futurenhs-deployments
+          path: futurenhs-deployments
+          #github deploy key?
```
alsuren, line 33: For reference, this is what the MHRA team are doing:
```
      - name: Clone Deployments repo
        uses: actions/checkout@v2
        with:
          repository: MHRA/deployments
          path: deployments
          token: ${{ secrets.DEPLOYMENTS_REPO_TOKEN }}
```

https://github.com/actions/checkout is the documentation for the  actions/checkout@v2 action. It documents path as 
```

    # Relative path under $GITHUB_WORKSPACE to place the repository
    path: ''
```
```

    # Personal access token (PAT) used to fetch the repository. The PAT is configured
    # with the local git config, which enables your scripts to run authenticated git
    # commands. The post-job step removes the PAT.
    #
    # We recommend using a service account with the least permissions necessary. Also
    # when generating a new PAT, select the least scopes necessary.
    #
    # [Learn more about creating and using encrypted secrets](https://help.github.com/en/actions/automating-your-workflow-with-github-actions/creating-and-using-encrypted-secrets)
    #
    # Default: ${{ github.token }}
    token: ''
```

frontend/.dockerignore
```diff
@@ -1,2 +1,3 @@
 node_modules
 npm-debug.log
+manifests/*
```
alsuren, line 3: Note for Jan: I was hoping that doing this would skip the build and test step (since the files won't have changed), but it doesn't, because the build agents don't have copies of the old layers. I think it's fine to leave for now, but might be worth revisiting. Maybe the MHRA approach of being able to manually commit to the argocd repo is better, because it allows you to act quicker.

.github/workflows/frontend-branch.yml
```diff
@@ -1,4 +1,4 @@
-name: frontend
+name: frontend-branch
```
frigus02, line 1: We should also update the path to this yaml file in the `on:` section below.

".github/workflows/frontend.yml" 👉 ".github/workflows/frontend-branch.yml"

.github/workflows/frontend-master.yml
```diff
@@ -0,0 +1,73 @@
...
+jobs:
+  ci:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone Platform repo
+        uses: actions/checkout@v2
+        with:
+          repository: FutureNHS/futurenhs-platform
+          path: futurenhs-platform
+          token: ${{secrets.DEPLOYMENT_TOKEN}}
```
frigus02, line 20: I think `repository` and `token` are not necessary here. We only need the path.

.github/workflows/frontend-master.yml
```diff
@@ -0,0 +1,73 @@
...
+          repository: FutureNHS/futurenhs-deployments
+          path: futurenhs-deployments
+          token: ${{secrets.DEPLOYMENT_TOKEN}}
+
+      - name: Copy deployment.yml
+        run: |
+          set -eux
+
+          cd $GITHUB_WORKSPACE/futurenhs-platform/frontend/manifests
+          cp deployment.yml kustomization.yaml $GITHUB_WORKSPACE/futurenhs-deployments/
```
frigus02, line 49: I might consider getting rid of this step and let the manifests live only in the deployments repository. But this is totally fine for now.

frontend/manifests/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- deployment.yml
+images:
+- digest: sha256:a9b186265ccaa78affc359970ae65a06386f4940e0927e5c965acde14f7b5bea
+  name: fnhsproduction.azurecr.io/frontend
```
frigus02, line 7: Should we remove the digest from here? We probably never update it in this repository.

.github/workflows/frontend-master.yml
```diff
@@ -0,0 +1,73 @@
...
+        uses: stefanprodan/kube-tools@v1
+        with:
+          kustomize: 3.5.5
+          command: |
+            set -eux
+            cd $GITHUB_WORKSPACE/futurenhs-deployments
+
+            kustomize edit set image $DIGEST
+            # TODO create service account, swap out email, update deployment_token secret
+            git config --local user.email "tracy.wu@red-badger.com"
```
frigus02, line 61: We can probably update this in a follow-up PR.

I also think we should document all the different secrets we create manually. The [docs/ci-setup.md](https://github.com/FutureNHS/futurenhs-platform/blob/master/docs/ci-setup.md) is probably a good place.

frontend/.dockerignore
frigus02: Yeah. In a personal project, I pulled the image first and told Docker to use it to find cache layers. It kind of works. But that's for the future when the build gets too slow.

I saw in MHRA that they're using  buildkit (see also https://github.com/MHRA/products/pull/721), which I think made the builds a lot faster. I'd try that first before trying to pull image layers.

frontend/manifests/kustomization.yaml
trcywu: We are leaving the step above where we cp this file into the deployment repo. Should we keep for now?

.github/workflows/frontend-master.yml
trcywu: yes hoping to do this in a follow-up PR. Have marked as TODO as to not forget.

frontend/manifests/kustomization.yaml
frigus02: Hm, it's not needed, is it? But if we plan to remove the entire file later on, then we might as well leave it as is for now.

trcywu: ok i will remove the digest

PR #29, opened 2020-06-25
suggested fixes for Potts

alsuren:
I was keeping notes by editing the readme while setting up the load balancer. Here's what I ended up with.

### DIFF
--- a/infrastructure/README.md
+++ b/infrastructure/README.md
@@ -112,6 +112,12 @@ as opposed to sharing a staging environment.
    terraform apply
    ```
 
+1. Change directory back to the root of the git repository:
+
+   ```bash
+   cd ../../..
+   ```
+
 1. Give the Kubernetes cluster permissions to pull images from our Docker registry.
 
    ```bash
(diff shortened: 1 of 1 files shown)


infrastructure/README.md
```diff
@@ -160,16 +166,16 @@ as opposed to sharing a staging environment.
...
 
 1. Spin up an Ingress Controller with your **name** and **email** as parameters. 
 
     This also creates a Load Balancer with a Public IP, creates a temporary domain, 
-    and registers a TLS Certificate with your temporary domain. 
+    and registers a TLS Certificate with your temporary domain (the domain must be across all azure users in europe)).
     
     ```bash
-    ./infrastructure/scripts/setup-dev-ingress.sh $FNHSNAME $FNHSEMAIL
+    ./infrastructure/scripts/setup-ingress.sh fnhs-dev-$FNHSNAME $FNHSEMAIL
```
cobwebfx, line 178: So this is done automatically in the setup-ingress.sh.  That way, we force fnhs-dev as the prefix.

infrastructure/README.md
```diff
@@ -160,16 +166,16 @@ as opposed to sharing a staging environment.
...
    ```bash
-   kubectl apply -f container-azm-ms-agentconfig.yaml
+   kubectl apply -f infrastructure/kubernetes/logging/container-azm-ms-agentconfig.yaml
    ```
 
 1. Spin up an Ingress Controller with your **name** and **email** as parameters. 
 
     This also creates a Load Balancer with a Public IP, creates a temporary domain, 
-    and registers a TLS Certificate with your temporary domain. 
+    and registers a TLS Certificate with your temporary domain (the domain must be across all azure users in europe)).
```
cobwebfx, line 175: What do you mean?

infrastructure/README.md
```diff
@@ -112,6 +112,12 @@ as opposed to sharing a staging environment.
    terraform apply
    ```
 
+1. Change directory back to the root of the git repository:
+
+   ```bash
+   cd ../../..
+   ```
```
cobwebfx, line 119: Not sure we need to hold the dev's hand here. 
We'd also need to remember to change this if we update the directory structure.  
@frigus02 mentioned in a different PR not to give the clone command because if a dev was to contribute to the project, they should really know how to clone.

infrastructure/README.md
alsuren: when I set it up, the domain david.westeurope.cloudapp.azure.com was already taken, so it failed.

frigus02: I think that's missing a "unique" in: the domain must be **unique** across...

PR #31, opened 2020-06-29
Cypress/BrowserStack spike

declanslevin:
## Name of story / Ticket #
[Spike: website automation testing](https://airtable.com/tblgtzV3sSwquTkIa/viw8ZHoNqvSvl62lt/recq5ANpYqkTWavcC)

## Acceptance Criteria

Spike different ways to test our website in browsers in an automated fashion.

- Cypress for Chrome, Edge
- Selenium/Browserstack for Internet Explorer

## Pre-review checklist:

- [x] Documentation


frigus02:
Nice. You probably want to create a `.gitignore` in the `test/` folder and put `node_modules/` inside. We don't need to commit the node modules.

Now that the node modules are already committed, it's a 2 step process:
1. Create `.gitignore` file, delete node modules and commit
2. Install node modules again (now git should not show these as changes anymore)

test/node_modules/.bin/_mocha
```diff
@@ -0,0 +1 @@
+../mocha/bin/_mocha
```
alsuren, line 1: We should probably add `node_modules` and `.DS_Store` to our .gitignore.

cobwebfx:
Overall I think this is good but is there a particular reason for using NPM over Yarn?

test/README.md
```diff
@@ -0,0 +1,45 @@
...
+## Cypress
+
+To run Cypress locally, cd into the cypress folder and execute `npm run cypress open`.
+
+This will open the test runner. You can then select a test you want to run and which browser you would like to run it in.
+
+## BrowserStack
+
+Before running any tests on BrowserStack you need to launch the BrowserStack Local executable.
+From the command line, cd into the browserstack folder and execute `./BrowserStackLocal --key {accesskey} --local-identifier bslocal1`.
```
alsuren, line 29: I'm a bit uncomfortable about checking a closed-source OSX-only binary into the git repo. Better to instruct people to `brew cask install browserstacklocal`, and remove the BrowserStackLocal binary file from the repo.

test/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+### Cypress
+
+To run Cypress on BrowserStack, duplicate and edit the browserstack.json example, adding your BrowserStack username and access key.
+
+Then execute `browserstack-cypress run`.
```
frigus02, line 35: For this to work you need to install the [browserstack cypress cli](https://github.com/browserstack/browserstack-cypress-cli) first.

test/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+### Selenium
+
+To run the Selenium test on BrowserStack, execute `node browserstack_local_test.js`.
```
frigus02, line 39: This doesn't work for me. I created the `.env` file inside the `test/` folder. This means I need to run it from there. So I need to run `node browserstack/browserstack_local_test.js`. However I then get the following error:

> WebDriverError: [browserstack.local] is set to true but local testing through BrowserStack is not connected.

I do have BrowserStackLocal running, though. I started it with the command mentioned above.

test/README.md
declanslevin: I missed this as a dependency so hopefully it should be installed now when running yarn.

test/README.md
declanslevin: I ran into a similar issue after updating everything so I removed the 'local-identifier' key from the command and it seems to be working properly now for me - I'll need to look into this more as it might present problems when trying to execute multiple testrunners in parallel

test/README.md
```diff
@@ -0,0 +1,51 @@
+# Automated Testing
+
```
cobwebfx, line 2: Would it be worth listing prerequisites, ie. everything that needs to be installed for this to run properly? If you don't have node, npm, yarn or chrome installed then I assume it won't work? It might be worth listing those dependencies, and explaining how to install them. I'm probably going overboard with node and npm, but the other 2 might not be obvious. Unless you disagree? We have it in the Infra Readme.

test/cypress/support/index.js
```diff
@@ -0,0 +1,20 @@
...
+// You can change the location of this file or turn off
+// automatically serving support files with the
+// 'supportFile' configuration option.
+//
+// You can read more here:
+// https://on.cypress.io/configuration
+// ***********************************************************
+
+// Import commands.js using ES2015 syntax:
+import './commands'
```
cobwebfx, line 17: I don't understand why we're importing a file that's completely commented out?

test/README.md
declanslevin: I agree we probably shouldn't assume future users will automatically have yarn etc. installed, I'll update the Readme

test/cypress/support/index.js
declanslevin: This is Cypress boilerplate which I haven't started making use of yet, but likely will going forward once things get a bit more complex.

cobwebfx: :thumbsup:

test/browserstack/browserstack_local_test.js
```diff
@@ -0,0 +1,43 @@
...
+  .usingServer(browserstackURL)
+  .withCapabilities(capabilities)
+  .build();
+
+driver.get("http://127.0.0.1:3000").then(function () {
+  driver.findElement(webdriver.By.css("h1")).then((h1) => {
+    h1.getText()
+      .then((text) => {
+        console.log(text);
+        assert.equal(text, "[Your Name]");
```
alsuren, line 35: looks like this will break when we merge with master because of 21c3535c017235ed9188e2996e9d7c939ea106c4

we can either: 
a) merge this PR into master as-is and make a separate PR to fix it
b) rebase this PR on top of master and fix the tests now

I would be in favour of (a), because I prefer short-lived branches and there is still value in having a set of failing cypress tests in master, because it still lets us think about how to integrate them into our deployment pipeline. Once we have the tests integrated into our deployment pipeline, we can start to prefer (b).

PR #34, opened 2020-06-30
Linkerd mobbing

frigus02:
- Linkerd proxy for frontend
- Linkerd proxy for hello-world
- Argo CD install script fix
- Ingress namespace fix


frigus02:
I would approve if I could. But I created this PR, so I can't. 🙂

PR #41, opened 2020-07-01
Client-side Instrumentation

matt-paul:
[ticket](https://airtable.com/tblgtzV3sSwquTkIa/viw8ZHoNqvSvl62lt/recVrr8dSCej1hevg?blocks=hide)

Notes:
 - Because Next.js makes use of Server Side Rendering, we have to make use of [this](https://github.com/goenning/next-applicationinsights#readme) library, which is built on top of the [official Microsoft library](https://github.com/microsoft/ApplicationInsights-JS)

- Page views, dependency calls and exceptions are tracked.

- These can be viewed within [Azure Application Insights](https://portal.azure.com/#blade/HubsExtension/BrowseResource/resourceType/microsoft.insights%2Fcomponents). 

Screenshot 2020-07-02 at 10 24 28

### DIFF
--- a/frontend/manifests/deployment.yaml
+++ b/frontend/manifests/deployment.yaml
@@ -15,6 +15,9 @@ spec:
       containers:
         - name: frontend
           image: fnhsproduction.azurecr.io/frontend
+          env:
+          - name: INSTRUMENTATION_KEY
+            value: ""
           ports:
             - containerPort: 3000
           resources:
--- a/frontend/pages/_app.tsx
+++ b/frontend/pages/_app.tsx
@@ -14,6 +14,6 @@ class MyApp extends App {
 }
 
 export default withApplicationInsights({
-  instrumentationKey: process.env.INSTRUMENTATION_KEY,
-  isEnabled: process.env.NODE_ENV === 'production'
+  instrumentationKey: "a2fa06cc-a1de-4efd-9096-9f99aa23a01c",
+  isEnabled: true //process.env.NODE_ENV === 'production'
 })(MyApp)
--- a/frontend/pages/_app.tsx
+++ b/frontend/pages/_app.tsx
@@ -1,6 +1,19 @@
 import React from "react";
-import { AppProps } from "next/app";
+import App from "next/app";
+import { withApplicationInsights } from 'next-applicationinsights';
 
-export default function App({ Component, pageProps }: AppProps) {
-  return <Component {...pageProps} />;
+class MyApp extends App {
+  render() {
+    const { Component, pageProps } = this.props
+
+    return (
+
+      <Component {...pageProps} />
+    )
+  }
 }
+
+export default withApplicationInsights({
+  instrumentationKey: '0b902312-dbaa-4ac0-bd21-029752baeb37',
+  isEnabled: true //process.env.NODE_ENV === 'production'
+})(MyApp)
--- a/frontend/pages/_app.tsx
+++ b/frontend/pages/_app.tsx
@@ -7,7 +7,6 @@ class MyApp extends App {
     const { Component, pageProps } = this.props
 
     return (
-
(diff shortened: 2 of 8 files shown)


frontend/pages/_app.tsx
```diff
@@ -1,6 +1,20 @@
...
+    return (
+
+      <Component {...pageProps} />
+    )
+  }
 }
+
+export default withApplicationInsights({
+  instrumentationKey: process.env.INSTRUMENTATION_KEY,
+  isEnabled: process.env.NODE_ENV === 'production'
```
cobwebfx, line 18: Should the word 'production' maybe be a const in some sort of config section in the codebase?

frigus02: `production` is a fairly standard thing for the `NODE_ENV` environment variable. In fact, I think Next.js sets this or expects it to be `production`. So I think this is fine.

frontend/manifests/deployment.yaml
```diff
@@ -15,6 +15,9 @@ spec:
       containers:
         - name: frontend
           image: fnhsproduction.azurecr.io/frontend
+          env:
+          - name: INSTRUMENTATION_KEY
+            value: ""
```
frigus02, line 20: We need to actually specify the instrumentation key here. Otherwise there don't be one when it's deployed to production.

This is really annoying now, because that key is different for every environment. Oh god. I'm starting to think we do need different folders or branches in the deployments repository.

frontend/pages/_app.tsx
```diff
@@ -1,6 +1,20 @@
...
-  return <Component {...pageProps} />;
+class MyApp extends App {
+  render() {
+    const { Component, pageProps } = this.props
+
+    return (
+
+      <Component {...pageProps} />
+    )
+  }
```
frigus02, line 13: Is this change of the `App` component necessary? Or does this work with the function component we had before? It doesn't look like the functionality is any different.

frontend/manifests/deployment.yaml
matt-paul: Yeah i'm a bit confused by the whole thing at the moment, we are at a weird stage 🤔

frontend/pages/_app.tsx
cobwebfx: That's fine.  I approved anyway.  I was just thinking about plain text variables being used in the codebase.  For instance, if the word "production" were to be used again in the codebase, accessing a const would be safer than typing it again.

frontend/pages/_app.tsx
matt-paul: You are right, at one stage it seemed like it would need to change (for a reason lost in time), but i agree it's nice to stick to function components

frontend/manifests/deployment.yaml
frigus02: Yeah, we need to properly solve that. It doesn't hold back this PR, though. You can just leave it blank for now.

frontend/pages/_app.tsx
matt-paul: Hmm, so actually, the reason I stuck with this (this was done about 10 days ago) was that the function 'withApplicationInsights' seems to expect it to be a class extending App

matt-paul: `
Argument of type '({ Component, pageProps }: AppPropsType<Router, {}>) => Element' is not assignable to parameter of type 'typeof App'.
  Type '({ Component, pageProps }: AppPropsType<Router, {}>) => Element' is missing the following properties from type 'typeof App': origGetInitialProps, getInitialPropsts(2345)
`

frigus02: Okay, fine. I guess there is a different. Maybe it's just a too specific type in the Next.js app insights lib.

PR #54, opened 2020-07-07
Use a 'Deploy Key' to clone the 'Deployments' repo

matt-paul:
sergey-raikin-gvU2BW7EXQE-unsplash

[Ticket](https://airtable.com/tblgtzV3sSwquTkIa/viw8ZHoNqvSvl62lt/recdG8LU95dtv2g1R?blocks=hide)

## Acceptance Criteria

- [x] Having removed the current token and replaced with the 'Deploy Key', we are successfully able to clone the 'Deployments repository from CI.

## Things
- The deploy key is an SSH key generated using `ssh-keygen -t rsa -b 4096 -C "futurenhs-devs@red-badger.com"`
- The public part of the key is added to the 'Deployments' repo (settings/deploy-keys).
- The private part of the key is added to the ‘Platform’ repo as a secret (settings/secrets), and is used within the ‘Platform’ workflow to clone the ‘Deployments’ repo.  
- Once cloned, the ‘Deployments’ repo is then be updated with the new disc image tag(s).

Once complete, the 'DEPLOYMENT_TOKEN' can be removed from settings/secrets

## Testing information

Good question - i guess, as long as a master build is able to clone the 'Deployments' repo succesfully then we have succeeded. When in development, this was tested by changing master workflows to run on PR update rather than push to master.

### DIFF
--- a/.github/workflows/frontend-master.yaml
+++ b/.github/workflows/frontend-master.yaml
@@ -37,7 +37,7 @@ jobs:
         with:
           repository: FutureNHS/futurenhs-deployments
           path: futurenhs-deployments
-          token: ${{secrets.DEPLOYMENT_TOKEN}}
+          ssh-key: ${{ secrets.DEPLOY_KEY }}
 
       - name: Copy manifests
         run: |
@@ -56,9 +56,6 @@ jobs:
             cd $GITHUB_WORKSPACE/futurenhs-deployments/frontend
 
             kustomize edit set image $DIGEST
-            # TODO create service account, swap out email, update deployment_token secret
-            git config --local user.email "tracy.wu@red-badger.com"
-            git config --local user.name "FutureNHS CI/CD"
             git add -A
             git diff-index --quiet HEAD || git commit -am "CI: Update production image for ${TAG}"
 
--- a/.github/workflows/frontend-master.yaml
+++ b/.github/workflows/frontend-master.yaml
@@ -1,7 +1,8 @@
 name: frontend-master
 
 on:
-  push:
+  # push:
+  pull_request:
     branches:
       - master
     paths:
--- a/.github/workflows/frontend-master.yaml
+++ b/.github/workflows/frontend-master.yaml
@@ -57,6 +57,10 @@ jobs:
             cd $GITHUB_WORKSPACE/futurenhs-deployments/frontend
 
             kustomize edit set image $DIGEST
+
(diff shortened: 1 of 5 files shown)


.github/workflows/infrastructure-master.yaml
```diff
@@ -37,9 +38,6 @@ jobs:
           set -eux
           cd $GITHUB_WORKSPACE/futurenhs-deployments
 
-          # TODO create service account, swap out email, update deployment_token secret
-          git config --local user.email "tracy.wu@red-badger.com"
-          git config --local user.name "FutureNHS CI/CD"
```
alsuren, line 42: I'm not sure what `git log` will look like with these unset. Maybe try it out and if you don't like it, set this to futurenhs-devs@red-badger.com and "FutureNHS CI/CD" ?

.github/workflows/frontend-master.yaml
```diff
@@ -48,25 +48,27 @@ jobs:
...
-            git config --local user.email "tracy.wu@red-badger.com"
-            git config --local user.name "FutureNHS CI/CD"
-            git add -A
-            git diff-index --quiet HEAD || git commit -am "CI: Update production image for ${TAG}"
+        run: |
+          set -eux
+
+          cd $HOME
+          curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"  | bash
+          cd $GITHUB_WORKSPACE/futurenhs-deployments/frontend
```
alsuren, line 57: what was wrong with stefanprodan/kube-tools?

frigus02: Took us a while to figure that one out. This GitHub action doesn't do anything wrong. But it uses GitHub actions' feature of running in a Docker container.

GitHub actions doesn't properly specify environment variables and volume mounts in order to use saved SSH keys. This means git push fails because it can't find the SSH key.

matt-paul: We found ourselves inside a Docker container, when we needed access to the parent filesystem.

frigus02: I'm about to create an issue. Just not exactly sure where. I think github.com/actions/checkout could be the best place for visibility.

frigus02: For reference: https://github.com/actions/checkout/issues/297

I hope this is the right place to report this. I could not find any repository, which is about running a container action.

PR #56, opened 2020-07-08
make `yarn dev` proxy /hello/* to the hello-world service

alsuren:
also make the nginx ingress not rewrite the url when proxying

This is based on https://github.com/vercel/next.js/tree/master/examples/with-custom-reverse-proxy.

## Test:

- [ ] Tested locally

### DIFF
--- a/frontend/server.js
+++ b/frontend/server.js
@@ -1,0 +1,52 @@
+/// This is taken from the next.js reverse proxy example, and should
+/// only be used for local development.
+/// https://github.com/vercel/next.js/tree/master/examples/with-custom-reverse-proxy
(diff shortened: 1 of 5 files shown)


frontend/server.js
```diff
@@ -0,0 +1,53 @@
+/// This is taken from the next.js reverse proxy example, and should
```
trcywu, line 1: Are we keeping this as a .js file or .ts?

trcywu: ok nvm i get it now. it's not getting bundled with the next.js app and is stand alone so no need to be .ts

PR #58, opened 2020-07-08
Environment configs for all services

frigus02:
This PR creates a manifest structure like this for all services:

```
frontend/manifests/
  base/
    deployment.yaml
    kustomization.yaml
    namespace.yaml
    service.yaml
  dev-template/
    deployment.yaml
    kustomization.yaml
  production/
    deployment.yaml
    kustomization.yaml
```

The same kind of structure exists for Argo CD applications, so that these apps point to the correct overlay folder.

Based on the `dev-template` folders, a new script `create-dev-overlays.py` uses a list of developer overlays in `dev-overlays.json` to generate the actual overlay folders for everyone. You can generate them locally if you want to synchronize local changes to Argo CD. And it runs in master builds to push everyone's overlays to the deployments repository.

## Pre-review checklist:

- [x] Test locally
- [x] Documentation


infrastructure/kubernetes/argocd/apps-dev-jan/cert-manager.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: argoproj.io/v1alpha1
```
alsuren, line 1: maybe move this from 
infrastructure/kubernetes/argocd/apps-dev-jan/cert-manager.yaml
to
infrastructure/kubernetes/argocd/apps/dev-jan/cert-manager.yaml
for symmetry?

frigus02: I was just thinking that. I wanted to keep the folder structure flat. But that's a lost battle anyway. So yeah, maybe:

-  infrastructure/kubernetes/argocd-apps/dev-jan/cert-manager.yaml

😄

infrastructure/scripts/install-argo-cd.sh
```diff
@@ -4,23 +4,24 @@ set -eu
 # https://stackoverflow.com/a/1482133
 cd $(dirname "$0")
 
-CURRENT_CLUSTER=$(kubectl config current-context)
+NAME="${1:?"Please specify you environment name as the first parameter, e.g. dev-jane"}"
```
alsuren, line 7: in the readme you say `install-argo-cd.sh $FNHSNAME` but FNHSNAME=john. Maybe make the scripts accept both?

frigus02: Ahh. I did not test that, yet. Hm, what is a nice way to do that? Change the README to say `install-argo-cd.sh dev-$FNHSNAME` or somehow figure out the right thing to do in the script?

infrastructure/scripts/create-dev-overlays.py
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env python3
```
alsuren, line 1: maybe call this `expand-dev-overlays.py` or something, or rename `infrastructure/dev-overlays.json` to `dev-overlay-variables.json`? From just the names, it sounds like this script might be responsible for creating the `dev-overlays.json` file.

frigus02:
I followed the readme step by step to create a new environment for myself. Everything worked except the applications didn't install because the folders don't exist in the deployments repo, yet. So that's going to start working once this is merged.

infrastructure/scripts/create-dev-overlays.py
frigus02: Ah good point. I like `create-dev-overlays.py` because it alighs with `create-dev-environment.sh`. But I can see that this could cause confusion.

I think I like your suggestion of renaming the JSON file.

PR #60, opened 2020-07-08
revert ingress change

alsuren:
Currently nginx ingress is broken for the hello-world service. This PR reverts it to the last known working version.


frigus02:
Hm, the current master seems to work for me. I just applied it to production as well and this works:

https://fnhs.westeurope.cloudapp.azure.com/hello/name

What's the problem you're having?

alsuren:
I just forgot to run `argocd app sync ingress --local ./infrastructure/kubernetes/ingress/dev/`. Nothing to see here.

PR #62, opened 2020-07-13
Synapse terraform

cobwebfx:
Create Synapse instance automatically via terraform

## Pre-review checklist:

- [x] Run Terraform in Dev


infrastructure/modules/platform/main.tf
```diff
@@ -97,3 +97,95 @@ resource "azurerm_application_insights" "app_insights" {
...
+resource "random_string" "server" {
+  length  = 15
+  special = false
+}
+
+data "azurerm_client_config" "current" {
+}
+
+resource "azurerm_key_vault" "secret_vault" {
+  name                = format("%s%s", "kv-fnhs", random_string.server.result)
```
alsuren, line 116: this is equivalent to:
```suggestion
  name                = "kv-fnhs${random_string.server.result}"
```

although I would suggest that `server` is not a great name for this, and it might benefit from including the environment name. Maybe:

```suggestion
  name                = "fnhs-keyvault-${var.environment}-${random_string.keyvault_suffix.result}"
```

infrastructure/modules/platform/main.tf
```diff
@@ -97,3 +97,95 @@ resource "azurerm_application_insights" "app_insights" {
   resource_group_name = azurerm_resource_group.platform.name
   application_type    = "other"
 }
+
+resource "random_password" "password" {
```
alsuren, line 101: I'm going to need another password as part of https://github.com/FutureNHS/futurenhs-platform/pull/61/files let's go with postgres_password and synapse_password or something?
```suggestion
resource "random_password" "synapse_password" {
```

infrastructure/README.md
```diff
@@ -86,6 +86,14 @@ as opposed to sharing a staging environment.
    ```bash
    cd infrastructure/environments/dev
    ```
+   
+1. If you intend to use Synapse, add your IP to the list in terraform.tfvars
+
+   ```hcl-terraform
+   ip_whitelist_insights = {
+     john = "123.45.678.90",
+   }
```
alsuren, line 95: ❤️  I might steal this trick for postgres. Can you think of a way to automate it?

infrastructure/modules/platform/main.tf
```diff
@@ -97,3 +97,95 @@ resource "azurerm_application_insights" "app_insights" {
...
+  name         = "synapse-password"
+  value        = random_password.password.result
+  key_vault_id = azurerm_key_vault.secret_vault.id
+
+  tags = {
+    environment = var.environment
+  }
+}
+
+resource "azurerm_mssql_server" "sql_server" {
```
alsuren, line 154: just trying to avoid a name-grab on `sql` when we are also expecting to have a postgres
```suggestion
resource "azurerm_mssql_server" "synapse_server" {
```

infrastructure/modules/platform/main.tf
```diff
@@ -97,3 +97,95 @@ resource "azurerm_application_insights" "app_insights" {
...
+    login_username = var.ad_username
+    object_id      = var.ad_object_id
+  }
+
+  tags = {
+    environment = var.environment
+  }
+}
+
+resource "azurerm_sql_database" "sql_dw" {
```
alsuren, line 171: I'm not used to the acronym dw for data warehouse
```suggestion
resource "azurerm_sql_database" "synapse_db" {
```
?

infrastructure/modules/platform/main.tf
cobwebfx: I wanted this also, but there is a character limit of 24 alphanumeric chars, and the name needs to be completely unique inside the whole of Azure.  This includes recently deleted names.
So if you created: 
fnhs-keyvault-dev-david${random_string.keyvault_suffix.result}
The random string would only be 1 character, which would give you a lower opportunity to get a unique name.  Also, "david" is 5 chars, so we don't have a way in terraform to chop part of a name off if someone has a name like "declan".

alsuren: (also avoiding `kv` because `kv` means `key-value` to me, so I would expect this to be a redis instance or something).

infrastructure/modules/platform/main.tf
cobwebfx: This can generate multiple passwords.  You have ask it to dump an array rather than a string.  I originally had it, but @frigus02 asked me to remove because we currently don't have more than one password.  I'm happy to pair with you, when you have to create a new password.

PR #64, opened 2020-07-14
Client side rendering

trcywu:
## Name of story / Ticket #
Link to Zeplin/design files

## Acceptance Criteria

This PR presents 2 examples side by side. Server side rendering vs client side
isomophic-fetch vs Axios

## Pre-review checklist:

- [ ] Tests

- [ ] Documentation

- [ ] Feature flag

## Testing information

- Test plan?

## Test:

- [ ] Tested locally 

- Platforms/Browsers: 

  - [ ] Internet Explorer 11
  
  - [ ] Edge
  
- Accessibility:

  - [ ] Screen Reader
  
  - [ ] Keyboard Navigation
  
  - [ ] Magnification
  
  - [ ] Contrast
  
  - [ ] Text size 200%
  
  - [ ] Landscape (Mobile)

### DIFF
--- a/frontend/pages/greetings/[name].tsx
+++ b/frontend/pages/greetings/[name].tsx
@@ -1,38 +1,58 @@
-import React, { useState } from "react";
+import React, { useState, useEffect } from "react";
 import Link from "next/link";
-import fetch from "node-fetch";
+import axios from "axios";
 
 import { GetServerSideProps } from "next";
+import { getGreeting } from "../../lib/greetings";
 
 export const getServerSideProps: GetServerSideProps = async (context) => {
-  const res = await fetch(
-    `http://hello-world.hello-world/hello/${context?.params?.name}`
-  );
-
-  console.log("*****8888", typeof window);
-
-  const greeting = await res.text();
+  const nameProp = context?.params?.name || "";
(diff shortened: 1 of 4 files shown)


frontend/pages/greetings/[name].tsx
```diff
@@ -1,38 +1,58 @@
...
+        <div>Server side rendering example</div>
         <div>Enter a name</div>
-
         <div>{greeting}</div>
         <input
           value={inputValue}
           onChange={(e) => setInputValue(e.target.value)}
         />
-
+        {/* shallow - Update the path of the current page without rerunning */}
```
Comment (alsuren) [inline frontend/pages/greetings/[name].tsx:55]: it's worth expanding on this with what the data flow looks like. I didn't believe how lightweight it was until you showed me the network tab.
```suggestion
        {/* shallow - Update the path of the current page and call getServerSideProps() on the server, then re-render on the client-side */}
```

It's also worth noting that given how lightweight the `<Link>`-based approach is, we should probably treat it as our default way of working.

Doing the fetch on the client side is probably a premature optimisation if it causes code duplication.

Comment (trcywu) [reply frontend/pages/greetings/[name].tsx:55]: yeah, next.js is really geared for server-side. client-side is so much more work.

PR #70, opened 2020-07-17
idempotent `CLEAN=localstate create-dev-environment.sh`

alsuren:
After applying this patch, `CLEAN=localstate ./create-dev-environment.sh david` will be idempotent: you can safely run it as many times as you want.

Also, `./create-dev-environment.sh david` will now also flag up if you have a .terraform dir lying around.

I have used environment variables because parsing flags is harder that I can be bothered with, and people should get used to using environment variables anyway, because that's how 12-factor apps are configured.

### DIFF
--- a/infrastructure/scripts/create-dev-environment.sh
+++ b/infrastructure/scripts/create-dev-environment.sh
@@ -10,12 +10,12 @@ if [ "$NAME" != "${NAME//[^a-z]/-}" ]; then
 fi
 
 # We might want to write a OVERWRITE=full version later that does a terraform destroy etc.
-OVERWRITE="${OVERWRITE:-nothing}"
-if [[ "$OVERWRITE" != "localfiles" && "$OVERWRITE" != "justdotterraform" && "$OVERWRITE" != "nothing" ]]; then
+OVERWRITE="${OVERWRITE:-forbidden}"
+if [[ "$OVERWRITE" != "localfiles" && "$OVERWRITE" != "justdotterraform" && "$OVERWRITE" != "forbidden" ]]; then
 	echo "Unexpected value for OVERWRITE environment variable: '$OVERWRITE'"
 	echo "Use OVERWRITE=localfiles to clean your local terraform.tfvars and .terraform."
 	echo "Use OVERWRITE=justdotterraform to clean just .terraform, but leave your terraform.tfvars alone."
-	echo "Use OVERWRITE=nothing (the default) if you are using a clean envionment."
+	echo "Use OVERWRITE=forbidden (the default) if you are using a clean envionment and want to abort if anything is unclean."
 	exit 1
 fi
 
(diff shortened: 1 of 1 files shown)


infrastructure/scripts/create-dev-environment.sh
```diff
@@ -58,18 +77,21 @@ setup_terraform() {
 		--account-key "$ACCESS_KEY" \
 		--output none
 
-	cat >"$DEV_CONFIG_FILE" <<EOF
+	if [ "$CLEAN" = "localstate" ]; then
```
frigus02, line 80: Should that be `if [ "$CLEAN" = "localstate" ] || [ ! -f "$DEV_CONFIG_FILE" ]`? Otherwise the file would not be created in `check` mode I think.

frigus02: Aaahh right. It should be `!=`. And then we don't need to check for an existing file.

frigus02: `OVERWRITE` might be clearer than `CLEAN`. I think I don't paricularly mind.

I still struggle with the different between `local` and `localstate` and have to think about this for a while before it makes sense to me. But that's alright. The help messages explains it very well.

infrastructure/scripts/create-dev-environment.sh
```diff
@@ -9,19 +9,44 @@ if [ "$NAME" != "${NAME//[^a-z]/-}" ]; then
...
+	elif [ "$OVERWRITE" = "justdotterraform" ]; then
+		rm -r $TERRAFORM_DIR
+	elif [ -f "$DEV_CONFIG_FILE" ]; then
+		echo "File $DEV_CONFIG_FILE already exists."
+		echo "If you want to initialize your environment again, please delete the file and rerun this script,"
+		echo "or run:"
+		echo ""
+		echo "    OVERWRITE=localfiles $0 $NAME"
+		echo "or:"
+		echo "    OVERWRITE=justdotterraform $0 $NAME"
```
frigus02, line 44: `justdotterraform` won't help in this case because it doesn't remove the config file. Should we move this suggestion to the `[ -d "$TERRAFORM_DIR" ]` block?

infrastructure/scripts/create-dev-environment.sh
```diff
@@ -58,17 +83,21 @@ setup_terraform() {
 		--account-key "$ACCESS_KEY" \
 		--output none
 
-	cat >"$DEV_CONFIG_FILE" <<EOF
+	if [ "$OVERWRITE" = "localfiles" ] || [ ! -f "$DEV_CONFIG_FILE" ]; then
```
frigus02, line 86: I think this `if` is not needed. All cases have been handled already above. Or am I missing something?

infrastructure/scripts/create-dev-environment.sh
frigus02: I misread this. Using justdotterraform actually does let the script run. It just means it won’t overwrite the tfvars file

infrastructure/scripts/create-dev-environment.sh
frigus02: Yep. See my comment above.

PR #71, opened 2020-07-20
Data map

simplybenuk:


frigus02:
Looks good to me. There is a typo in the folder name. I can fix that.

I think we should also draw this diagram in draw.io, so that everyone is able to edit it. But we can do that later on.

PR #72, opened 2020-07-22
Argocd manages branch clusters.

matt-paul:
Screenshot 2020-07-22 at 13 02 56

This PR primarily relates to setting up ArgoCD to manage dev clusters, and lays the groundwork to install Azure Service Operator and the PostgreSQL database which it will manage.

- [x] Update branch workflows to include clone / copy manifests/ deploy steps.
- [x] Add 'out of retries' message to master workflows.
- [x] Tweak 'Working with Sealed Secrets.md' for better readability. 
- [x] Fix secretKeyRef and update related Sealed Secret 
- [x] Update infrastucture README.
- [x] Add branch to dev-overlay variables 

## Testing
1. Update BRANCH variable in infrastructure/dev-overlay-variables.json to match this branch name.
2. Run ./install-argo script to update your local argo apps.
3. Run kubectl port-forward svc/argocd-server -n argocd 8080:443 to launch UI at localhost:8080.
4. Login using admin as user and set the passord to the name of the argocd server pod (kubectl get pods -n argocd to find it)
4. Verify in UI


infrastructure/scripts/install-argo-cd.sh
```diff
@@ -25,6 +25,9 @@ kustomize build ../kubernetes/argocd/install |
 	kubectl apply -n argocd -f -
 kubectl rollout status -n argocd deployment argocd-server
 
+echo "Expanding dev overlyas"
```
frigus02, line 28: Small typo 🔍 
```suggestion
echo "Expanding dev overlays"
```

frigus02: Also, now that we create the dev overlays automatically, we can remove line 34:
```
	echo "Did you run ./infrastructure/scripts/create-dev-overlays.py?"
```

.github/workflows/hello-world-branch.yaml
```diff
@@ -24,8 +27,58 @@ jobs:
...
+          make docker-push tag=${TAG}
+          DIGEST="$(docker inspect --format='{{index .RepoDigests 0}}' ${{ secrets.REGISTRY_LOGIN_SERVER }}/hello-world:${TAG})"
+          echo ::set-env name=TAG::$TAG
+          echo ::set-env name=DIGEST::$DIGEST
+          echo ::set-env name=DEPLOYMENTS_BRANCH::pr-$GITHUB_HEAD_REF
+      - name: Clone Deployments repo
+        uses: actions/checkout@v2
+        with:
+          repository: FutureNHS/futurenhs-deployments
+          ref: ${{ env.DEPLOYMENTS_BRANCH }}
```
frigus02, line 42: I'm not sure if this works. I think this may fails because the branch does not exist. In the infrastructure build the checkout happens manually. We probably need to do that here, too.

.github/workflows/frontend-branch.yaml
```diff
@@ -21,8 +24,68 @@ jobs:
...
+          make docker-push tag=${TAG}
+          DIGEST="$(docker inspect --format='{{index .RepoDigests 0}}' ${{ secrets.REGISTRY_LOGIN_SERVER }}/frontend:${TAG})"
+          echo ::set-env name=TAG::$TAG
+          echo ::set-env name=DIGEST::$DIGEST
+          echo ::set-env name=DEPLOYMENTS_BRANCH::pr-$GITHUB_HEAD_REF
+      - name: Clone Deployments repo
+        uses: actions/checkout@v2
+        with:
+          repository: FutureNHS/futurenhs-deployments
+          ref: ${{ env.DEPLOYMENTS_BRANCH }}
```
frigus02, line 39: I'm not sure if this works. I think this may fails because the branch does not exist. In the infrastructure build the checkout happens manually. We probably need to do that here, too.

.github/workflows/frontend-branch.yaml
```diff
@@ -21,8 +24,68 @@ jobs:
...
+          ssh-key: ${{ secrets.DEPLOY_KEY }}
+      - name: Update image tag and test dev overlay creation
+        run: |
+          cd $HOME
+          curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh"  | bash
+
+          # it might actually be possible to kustomize the production version
+          # and let the other overlays pick up the change.
+          # TODO: would be nice if the deployments repo followed the same structure
+          # as the platform repo, but only contained manifests.
```
frigus02, line 50: I'm not sure if I agree, that having the exact same structure as the platform repo would be a good thing. I think it would make the deployments repo harder to read. 🤔

.github/workflows/infrastructure-branch.yaml
```diff
@@ -12,22 +12,81 @@ jobs:
...
+        working-directory: futurenhs-platform/infrastructure/environments/dev
       - name: Validate Terraform of production environment
         run: |
           terraform init -backend=false
           terraform validate
-        working-directory: infrastructure/environments/production
+        working-directory: futurenhs-platform/infrastructure/environments/production
       - name: Test dev overlay creation
-        run: infrastructure/scripts/create-dev-overlays.py
+        run: $GITHUB_WORKSPACE/futurenhs-platform/infrastructure/scripts/create-dev-overlays.py
```
frigus02, line 36: Also create dev overlays as part of the "Copy manifests" step now. So I think we can remove the "Test dev overlay creation" step.

docs/working-with-sealed-secrets.md
```diff
@@ -12,7 +12,12 @@ az keyvault secret show --vault-name "fnhs-shared-dev" --name "sealed-secret-yam
 
 Create a yaml-encoded Kubernetes Secret: (note use of `--dry-run` - this is just a local file and should be deleted afterwards)
 ```
-$ kubectl create secret generic favouritecuisine --dry-run --from-literal=favouritecuisine=indian -o yaml > favourite-cuisine.yaml
+$ kubectl create secret generic favouritecuisine \
+   --dry-run \
+   --namespace=hello-world \
+   --from-literal=favouritecuisine=spanish \
+   -o yaml \
+   | kubeseal --format yaml > favouritecuisine-sealed.yaml
```
frigus02, line 20: I like this change. We should now remove the following paragraph, because it's not longer valid.

frontend/Makefile
```diff
@@ -0,0 +1,28 @@
...
+
+.PHONY: docker-build
+docker-build: ## Build and tag Docker image
+	DOCKER_BUILDKIT=1 docker build \
+		--progress plain \
+		--tag $(app) \
+		--tag $(image) \
+		--tag $(image):$(tag) \
+		.
+.PHONY: docker-run
```
frigus02, line 17: There should probably be an empty line between line 16 and 17 🔍 ✨

.github/workflows/frontend-branch.yaml
matt-paul: one for @alsuren

alsuren: github and vscode both do a reasonable job of flattening out directories that are empty, and I think that the symmetry would be valuable in reducing the number of translations we need to do in our heads when dealing with infrastructure.

I guess I can try writing the documentation for this, and see if anything jumps out for simplification.

Currently we have (from `git ls-files | grep yaml`)
```
# a bunch of github actions descriptions
.github/workflows/*.yaml

# Descriptions of apps that we wrote. The filename mapping from platform to deployments for these files is `s:manifests/::`:
# these are also deployed by the {frontend|hello-world}-{branch|master} jobs, rather than the infrastructure-* jobs, so they might take ~10 minutes to actually get pushed.

# standard-ish service definitions that are deployed by argocd
{frontend|hello-world}/manifests/base/*.yaml
# thin patches over the base definitions
{frontend|hello-world}/manifests/production/*.yaml
# crazy custom overrides that mostly consists of patches with $NAME substitutions
{frontend|hello-world}/manifests/dev-template/*.yaml
# expanded templates (not checked into git on the platform repo, but are in the deployments repo)
{frontend|hello-world}/manifests/dev-*/*.yaml

# Bootstrapping for argocd. These are only used locally.
infrastructure/kubernetes/argocd/install/*.yaml

# argocd "Application" descriptions. These don't get pushed to the deployments repo at the moment (but maybe this will change later?). These are only used locally.
infrastructure/kubernetes/argocd/apps/production/*.yaml

# When viewing/editing these files, and trying to work out out the mapping between `production/$appname.yaml` and the filename they correspond to, there is no deterministic reverse filename mapping, so you have to:
# * know by magic (or reading the corresponding production/*.yaml) that you need to be looking in the deployments repo
# * find the appropriate file in the deployments repo, and trace backwards to where it lives in the platform repo (from the github actions cp commands?), or just grep for the name of the app, and hope that you get it right.
infrastructure/kubernetes/argocd/apps/dev-template/*.yaml
infrastructure/kubernetes/argocd/apps/dev-*/*.yaml

# a couple of random files that we install by hand in the README.md script
infrastructure/kubernetes/logging/container-azm-ms-agentconfig.yaml
infrastructure/kubernetes/sealed-secrets/controller.yaml


# If the name is *not* argocd or logging or sealed-secrets then the rest of the files are descriptions of services that we have not written ourselves, and will end up in the deployment repo. The filename mapping seems to be s:infrastructure/kubernetes/::
# These are all pushed to the deployments repo and deployed by argocd?
infrastructure/kubernetes/{azure-service-operator,cert-manager,ingress}/production/*.yaml
infrastructure/kubernetes/*/dev-template/*.yaml
infrastructure/kubernetes/*/dev-*/*.yaml
```

There are a couple of TODOs that jump out at me:

- [ ] make a simple reversible mapping between the names of files in the platform repo and their names in the deployments repo. I propose the identity mapping.
- [ ] separate manually applied cluster-bootstrapping files into their own hierarchy
- [ ] collapse a bunch of the README steps into scripts.

frigus02: > github and vscode both do a reasonable job of flattening out directories that are empty

Hm, that's true.

Yeah, maybe having the same structure can work. Maybe it's just me and what I'm used to, that makes me think the current structure of the deployments repo is better. Happy to give this a try.

PR #75, opened 2020-07-23
Set up Azure Service Operator & PostgreSQL

matt-paul:
Screenshot 2020-07-23 at 14 45 17

[ticket](https://airtable.com/tblgtzV3sSwquTkIa/viw8ZHoNqvSvl62lt/rec7odcGP0zFuV7kH?blocks=hide)

## Overview
This PR includes Azure Service Operator for Kubernetes, and our Postgresql db.

[Azure Service Operator for Kubernetes](https://github.com/Azure/azure-service-operator) provisions resources outside of the cluster, and connects our applications to them from within Kubernetes. As well as PostgresSQL, it also supports Azure Key Vault amongst others, so we may make use of it in other places too.

- [x] Add Azure Service Operator
- [x] Add postgresql
- [x] Add health checks for postgresql resources (infrastructure/kubernetes/argocd/install/patch-argocd-cm.yaml)

## Testing
 - [x] Verify that the db is in development cluster as expected.
 - [ ] Verify that health check script works.


frigus02:
I think this should tell Argo CD how to check if the Postgres resources are health: https://github.com/FutureNHS/futurenhs-platform/commit/29d538b38a4e3937277b5125faf11854e7d81251

hello-world/manifests/base/postgresql.yaml
```diff
@@ -0,0 +1,66 @@
+apiVersion: azure.microsoft.com/v1alpha1
+kind: PostgreSQLServer
+metadata:
+  name: postgresqlserver-sample
+  namespace: hello-world
+spec:
+  location: westeurope
+  resourceGroup: platform-dev-david
```
alsuren, line 8: :/ I noticed this when doing my spike

infrastructure/kubernetes/argocd/install/patch-argocd-cm.yaml
```diff
@@ -0,0 +1,74 @@
...
+  name: argocd-cm
+data:
+  resource.customizations: |
+    azure.microsoft.com/PostgreSQLServer:
+      health.lua: |
+        hs = {}
+        if obj.status ~= nil then
+          hs.message = obj.status.message
+          if obj.status.failedProvisioning then
+            hs.status = "Degraded"
```
alsuren, line 13: ❤️

infrastructure/kubernetes/azure-service-operator/production/kustomization.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - namespace.yaml
+  - operator.yaml
+  # - azure-service-operator-secrets-sealed.yaml
```
alsuren, line 6: we will need to generate one of these for production as well. Happy to pair on this

frigus02:
We still need to sort this out:

> dev-david-operator-creds (service principal) - apparently the client secrets expire after some time? 16/7/2021 apparently, according to
https://portal.azure.com/#blade/Microsoft_AAD_RegisteredApps/ApplicationMenuBlade/Credentials/appId/829616f1-a89f-453a-90ca-5f0a30dcfa00/isMSAApp/

Service principal client secrets expire after 1 year. And I'm not aware of a way to change this.

I think the way to go is to use a managed identity instead of a service principal: https://github.com/Azure/azure-service-operator/blob/master/docs/howto/managedidentity.md

That requires us to install https://github.com/Azure/aad-pod-identity/. This sounds like a nice way to communicate with any kind of Azure resource. It's basically GCP WorkloadIdentity for Azure.

I'm not sure if we should do that in this PR or in a follow-up. It's probably nicer to do it step by step.

hello-world/manifests/base/postgresql.yaml
frigus02: I think this means we leave this out out `base` and only add it in `dev-template` using `platform-dev-$NAME`

infrastructure/kubernetes/argocd/install/patch-argocd-cm.yaml
frigus02: I didn't actually test this 🙈. Did this work, Matt?

hello-world/manifests/base/postgresql.yaml
matt-paul: doh

alsuren:
> I'm not sure if we should do that in this PR or in a follow-up. It's probably nicer to do it step by step.

let's do that as a separate PR. I guess we can track these types of TODOs on the ticket, and then decide which of them we want to make backlog tickets for and which we want to solve immediately?

frigus02:
Sounds good to me. This particular one, I would do right after this PR is merged 🙂

hello-world/manifests/base/postgresql.yaml
matt-paul: Removing this file - will be back in a future PR (as discussed)

frigus02:
Azure actually seems to recommend using AAD Pod Identity, so that does seem like the way to go: https://docs.microsoft.com/en-us/azure/aks/operator-best-practices-identity#use-pod-identities

hello-world/manifests/base/postgresql.yaml
```diff
@@ -0,0 +1,62 @@
+apiVersion: azure.microsoft.com/v1alpha1
```
matt-paul, line 1: Rather than defining the db server within Terraform, Azure Service Operator allows us to define it in our manifests

frigus02:
I'm going to close this. It's superseeded by #77

PR #76, opened 2020-07-23
Share Synapse between dev environments

frigus02:
The Synapse SQL DataWarehouse is super expensive compared to the rest of the infrastructure. This moves Synapse to the shared dev environment, which means we end up with a single instance in dev and a single instance in production.

We can revisit this later on and potentially move it back to the individual environments when the need arises.

---

Notes:
- The change in the infrastructure readme is unrelated. I thought it needs updating and ran Prettier on it. Since it did change quite a bit, I thought I might as well commit that.


infrastructure/scripts/create-dev-environment.sh
```diff
@@ -62,7 +62,6 @@ setup_terraform() {
 resource_group_name="$RESOURCE_GROUP_NAME"
 storage_account_name="$STORAGE_ACCOUNT_NAME"
 USERNAME="$NAME"
-ip_whitelist_insights={}
```
alsuren, line 65: now that this has been moved, do we need to deal with it in `infrastructure/scripts/create-shared-dev-environment.sh` ?

frigus02: Yeah. Similar to how it works for production. I'm not sure how we want to deal with this.

- Should we commit a terraform.tfvars file? I think that would be annoying
- Should we add an example file? That can be inferred from the variables.tf, so it seems like unnecessary duplication.
- Should we just leave it as it is now and everyone needs to create the file manually?

frigus02:
> In general I would prefer that style commits happen as their own commit. In practice once we have linters in ci we should stop having this problem so ✅

Agree. I was lazy. Fixed it now.

infrastructure/scripts/create-dev-environment.sh
frigus02: Okay, I just now saw the script to create the shared dev environment. I will make it generate a terraform.tfvars file.

frigus02:
Done and applied

PR #78, opened 2020-07-23
Add AAD Pod Identity through Argo CD

frigus02:
Follow-up to #75 

This deploys AAD Pod Identity through Argo CD.

### DIFF
--- a/infrastructure/modules/platform/main.tf
+++ b/infrastructure/modules/platform/main.tf
@@ -79,6 +79,16 @@ resource "azurerm_kubernetes_cluster" "cluster" {
   }
 }
 
+# Give the cluster managed identity permissions to read and assign
+# managed identities. This is required for AAD Pod Identity.
+# See: https://github.com/Azure/aad-pod-identity/blob/master/docs/readmes/README.role-assignment.md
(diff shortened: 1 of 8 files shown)


infrastructure/modules/platform/main.tf
```diff
@@ -79,6 +79,16 @@ resource "azurerm_kubernetes_cluster" "cluster" {
   }
 }
 
+# Give the cluster managed identity permissions to read and assign
+# managed identities. This is required for AAD Pod Identity.
+# See: https://github.com/Azure/aad-pod-identity/blob/master/docs/readmes/README.role-assignment.md
+resource "azurerm_role_assignment" "managed_identity_operator_cluster" {
+  scope                            = azurerm_resource_group.platform.id
+  role_definition_name             = "Managed Identity Operator"
```
frigus02, line 87: We might also need the "Virtual Machine Contributor" role. But it's not mentioned in any other documentation. It might just be neceesary for aks-engine, which I think is not what we're using.

frigus02:
We thought we need this for #75. But since we create the DB through terraform now, the AAD Pod Identity is no longer needed. We may come back to it later.

PR #79, opened 2020-07-27
Automation environments

declanslevin:
Cypress is now set up to run in a local browser/BrowserStack and in various environments (local/dev/prod). 
The Selenium/Mocha test for Internet Explorer will currently only run on Browserstack, against all environments. 

_____________

Unfortunately, the browserstack-cypress-cli doesn't allow you set a custom config at runtime so we had to get a little hacky with a script to create the appropriate config file: 
[test/scripts/cypress-bs.sh](https://github.com/FutureNHS/futurenhs-platform/blob/automation-environments/test/scripts/cypress-bs.sh) 

Please have a look and let me know if you can think of a better way of doing this.


test/package.json
```diff
@@ -1,9 +1,13 @@
 {
   "scripts": {
-    "cypress": "cypress open",
-    "cypress-bs": "browserstack-cypress run",
+    "cypress:local": "cypress open",
+    "cypress:dev": "./scripts/make-dev-config-local.sh",
```
alsuren, line 4: The name of this script should probably be similar to that of `./scripts/cypress-bs.sh` since it does a similar job (makes a config *and* runs cypress against it). Maybe `./scripts/cypress.sh` would be good enough?

test/package.json
```diff
@@ -1,9 +1,13 @@
 {
   "scripts": {
-    "cypress": "cypress open",
-    "cypress-bs": "browserstack-cypress run",
+    "cypress:local": "cypress open",
```
alsuren, line 3: Need to be explicit about which config file we're using here.
```suggestion
    "cypress:local": "cypress open --config-file ./cypress-local.json",
```

test/scripts/make-dev-config-local.sh
```diff
@@ -0,0 +1,14 @@
...
+
+set -eu
+
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+
+NAME="${1:?"Please enter your name as first argument"}"
+
+cat $REPO_ROOT/test/cypress-prod.json \
+    | sed "s/fnhs.westeurope.cloudapp.azure.com/fnhs-dev-$NAME.westeurope.cloudapp.azure.com/" \
+    > $REPO_ROOT/test/cypress.json
```
alsuren, line 11: Can you call out to `test/scripts/make-dev-config.sh` here, to avoid repeating yourself?

test/package.json
frigus02: I read that and thought you mean bs = bulls**t 😄. I guess it's browserstack

alsuren: Maybe we should be nicer about them. They are letting us use their service for free. That said, I'm happy to pay into the acronym swear jar for BS.

test/package.json
```diff
@@ -1,9 +1,13 @@
...
-    "cypress-bs": "browserstack-cypress run",
+    "cypress:local": "cypress open --config-file ./cypress-local.json",
+    "cypress:dev": "./scripts/cypress.sh",
+    "cypress:prod": "cypress open --config-file ./cypress-prod.json",
+    "cypress-bs:local": "./scripts/cypress-bs.sh ./cypress-bs-local.json",
+    "cypress-bs:dev": "./scripts/cypress-bs.sh",
+    "cypress-bs:prod": "./scripts/cypress-bs.sh ./cypress-prod.json",
     "mocha": "mocha",
-    "mocha-bs": "mocha ./browserstack/browserstack_local_test.js",
+    "mocha-bs": "mocha ./browserstack/browserstack_local_test_mocha.js",
```
frigus02, line 10: Do I understand this correctly, that mocha is running on BrowserStack using Selenium and can be used to test Internet Explorer?

I wonder if we should rename these to their browser targets, instead of the library they use: "yarn internet-explorer" and "yarn sensible-browsers" 🙂.

declanslevin: Pretty much, Selenium webdriver is used to access the elements in the page and the test assertions are performed with Mocha (and Chai). In theory the webdriver tests should actually work with any of the major modern browsers, but as Cypress is a faster and more reliable option we're not bothering to use Selenium for those.

I think renaming the 'browserstack' folders/tests is logical, but otherwise I think I'll leave the cypress folders/scripts as they are for now

frigus02: Yep, I agree. I wasn't too sure about the naming. It's a good point that the selenium tests actually work for all browsers.

declanslevin:
> Seems fine to. Do we need to update the readme to explain how to run the tests?

The Readme has been updated to reflect the current state of test running

test/README.md
```diff
@@ -4,50 +4,58 @@
...
 
-#### Selenium
+#### Internet Explorer
 
-The 'Selenium' test is essentially the same as the Mocha test, just without the Mocha syntax and formatting.
+To run tests, execute the desired command from the `/test/` directory:
 
-To run the Selenium test on BrowserStack, from the /test/ directory, execute `node /browserstack/browserstack_local_test.js`.
+- a local server: `yarn internet-explorer-bs:local`.
+- a dev cluster: `yarn internet-explorer-bs:dev <dev cluster name>` i.e. `yarn internet-explorer:dev foo`.
```
frigus02, line 60:
```suggestion
- a dev cluster: `yarn internet-explorer-bs:dev <dev cluster name>` i.e. `yarn internet-explorer-bs:dev foo`.
```

test/README.md
```diff
@@ -4,50 +4,58 @@
...
 
-### Mocha
+##### Run tests
 
-Mocha tests run using the Selenium webdriver.
+To run tests, execute the desired command from the `/test/` directory:
 
-To run the Mocha test on BrowserStack, execute `yarn mocha-bs`.
+- a local server: `yarn cypress-bs:local`.
+- a dev cluster: `yarn cypress-bs:dev <dev cluster name>` i.e. `yarn cypress:dev foo`.
```
frigus02, line 52: I ran this and got the following error:

```
[7/30/2020, 9:21:51 AM] - error: Build creation failed. Error: [local] is set to true but local testing through BrowserStack is not connected. To resolve the issue or to get more info visit - https://www.browserstack.com/question/779
```

I had to edit my browserstack.json file and change `local` to `false`. Then it worked.

Is that the correct thing to do?

declanslevin: Did you run `yarn local` to run the BrowserStackLocal executable first?

frigus02: Yes. That was all running.

I created a `.env` file based on the `.env.example`. Then I ran `yarn local` and kept that running.

I created a `browserstack.json` file from the `browserstack.json.example` and only changed the credentials at the top. Then I ran `yarn cypress-bs:dev tracy` and got the error.

declanslevin: I think I see the issue, I forgot to remove `"local_identifier": "bslocal1"` from `browserstack.json.example` (after removing it from the `local-bs-connect.sh`), does it work if you remove that?

It's only necessary to use `local_identifier` for running the selenium and cypress tests concurrently.

frigus02: Yes, that works 🎉

PR #81, opened 2020-07-29
Linting for Frontend

matt-paul:
Screenshot 2020-07-29 at 16 18 11

[Ticket](https://airtable.com/tblgtzV3sSwquTkIa/viw8ZHoNqvSvl62lt/recUEatSIsuRHWDXk?blocks=hide)

-Prettier for formatting (js/ts, tsx/jsx, md, yaml, graphql)
-Eslint for code quality (largely setup previously).

## Acceptance Criteria
- [x] Prettier runs in CI.
- [x] Prettier usable locally.
- [x] Eslint runs in CI.
- [x] Docs written for editor integration.
- [x] Husky pre-commit hook for Prettier and Eslint.
- [x] Ugly existing code prettified.
- [x] Eslint issues fixed for existing code. 

## Testing
- Format your code in the sloppiest way you can imagine.  Commit it and verify it fixes itself.
- Then verify that Prettier and ESlint also run succesffuly in CI

This is an example of what we are looking for in CI.  This was absolutely, 100% a deliberate mistake.  😅🤞
Screenshot 2020-07-30 at 12 15 17


.github/workflows/frontend-branch.yaml
```diff
@@ -9,7 +9,31 @@ on:
...
+  prettier:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.head_ref }}
+
+      - name: Prettify code
+        uses: creyD/prettier_action@v2.2
```
frigus02, line 33: I'm not sure about running Prettier with a special action. That means it might not use the same version we have defined in our package.json.

Maybe we can move this into the `eslint` step and run `yarn prettier --check ...`

.github/workflows/frontend-branch.yaml
```diff
@@ -9,7 +9,31 @@ on:
       - .github/workflows/frontend-branch.yaml
 
 jobs:
-  ci:
+  eslint:
```
frigus02, line 12: What's the reason for using a separate step? Asking out of curoisity. It repeats the checkout and yarn install step.

Does this have performance benefits because it runs in parallel?

frontend/.prettierignore
```diff
@@ -0,0 +1,7 @@
+.next
+node_modules
+public
```
frigus02, line 3: What's the reason for ignoring `public/`?

matt-paul: good spot, i don't think that should be there

.github/workflows/frontend-branch.yaml
matt-paul: I guess the main reason is because i prefer the UX when doing this.  It does run in parrallel too, but i'm not sure there is a massive benefit in terms of time.  Happy either way

frigus02: Seems fine to me. We just don't want it to take longer because it's doing some work twice 🙂

.github/workflows/frontend-branch.yaml
matt-paul: Ok cool.  For some reason, i thought we had a preference for using off the shelf actions! Happy to change.

frigus02: For me it's actually quite the opposite. I like to have a shell script, that runs in CI, that I can also run locally if I want to. But actions can make the CI workflows easier to understant. Both ways is probably fine.

Here the main reason is really using the same Prettier version. The extra action doesn't really matter. Prettier sometimes changes formatting, even for minor version updates. I think this is completely fine. But in does mean that our build could fail, just because the action decided to use a newer version of prettier, which would be super annoying.

matt-paul: yeah that makes sense.

PR #83, opened 2020-07-29
Redirect login to private page

trcywu:
[Ticket]()

[Zeplin]()

## Acceptance Criteria

This ticket builds the below flow:

A user tries to access a private workspace URL http://127.0.0.1:4455/workspaces/private -> check if user is logged in by checking if ory_kratos_session exists. 

If it does -> render private page

if it does -> redirect to login page -> user logs in -> redirects back to private workspace <workspaces/private>

## Pre-review checklist:

- [ ] Tests

- [ ] Documentation

- [ ] Feature flag

## Testing information

- Test plan?

## Test:

- [ ] Tested locally 

- Platforms/Browsers: 

  - [ ] Internet Explorer 11
  
  - [ ] Edge
  
- Accessibility:

  - [ ] Screen Reader
  
  - [ ] Keyboard Navigation
  
  - [ ] Magnification
  
  - [ ] Contrast
  
  - [ ] Text size 200%
  
  - [ ] Landscape (Mobile)

### DIFF
--- a/frontend/utils/pages/auth.ts
+++ b/frontend/utils/pages/auth.ts
@@ -6,6 +6,13 @@ export const requireAuthentication = async (
   context: GetServerSidePropsContext
 ) => {
   const { ory_kratos_session } = cookies(context);
+
+  if (context.res && !ory_kratos_session) {
+    context.res.writeHead(302, {
+      Location: `/.ory/kratos/public/self-service/browser/flows/login?return_to=${context.req.url}`,
+    });
+    context.res.end();
+  }
   try {
     const sessionResponse = await axios.request({
       url: "http://127.0.0.1:4433/sessions/whoami",
@@ -17,11 +24,7 @@ export const requireAuthentication = async (
 
     return sessionResponse;
   } catch (error) {
-    if (context.res) {
-      context.res.writeHead(302, {
-        Location: `/.ory/kratos/public/self-service/browser/flows/login?return_to=${context.req.url}`,
-      });
-      context.res.end();
-    }
+    console.error(error);
+    throw error;
   }
 };
--- a/frontend/utils/pages/auth.ts
+++ b/frontend/utils/pages/auth.ts
@@ -1,0 +1,14 @@
+import cookies from "next-cookies";
+import { GetServerSidePropsContext } from "next";
+
+export const requireAuthentication = async (
+  context: GetServerSidePropsContext
+) => {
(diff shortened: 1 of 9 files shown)


frigus02:
The `return_to` query parameter is nice.

2 questions to my mind:

- Does that also work when the user is already on the site and clicks on the link to the private page?

- This is quite a bit of code. Does Next.js have a way of wrapping a page somehow to make it authenticated? Although I guess we could write a helper function and then every site page that requires authentication could look like this:

  ```ts
  export const getServerSideProps: GetServerSideProps = async (context) => {
    await requireAuthentication(context);
    // optional custom logic
    return { props: {} };
  };
  ```

trcywu:
Yep. Also works if user is already on the home page, there's no token cookie, they click on the private workspace link, which takes them to login, they login and then redirect to private workspace. So ultimately the end up on the private workspace.

Yeah i think it can eventually be extracted as a helper function. But the url's are all different at the moment.

frigus02:
Amazing.

I assume there is a way of extracting the URL from the context. Something like `context.req.path` could probably be used as the `return_to` parameter.

frigus02:
I'm not 100% sure if we would use the cookie to decide if the user is logged in. We might use a special header, which our NGINX load balancer propagates from a call to Kratos. But I don't think that would change much. We would just replace `cookies(context)` with a custom check in `context.req.headers` or someting like this.

I think that's. I don't think we need anything else.

You could check if `context.req.path` works, to make the authentication check not specific to the page. Just to be 100% sure it's going to work later on. I can't think of anything else we need here.

frontend/utils/pages/auth.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  if (context.res && !ory_kratos_session) {
+    context.res.writeHead(302, {
+      Location: `/.ory/kratos/public/self-service/browser/flows/login?return_to=${context.req.url}`,
+    });
+    context.res.end();
+  }
+  try {
+    const sessionResponse = await axios.request({
+      url: "http://127.0.0.1:4433/sessions/whoami",
```
alsuren, line 18: This can probably use a url that points to the server that's in the cluster (`http://kratos-public.kratos/sessions/whoami`). If you want to continue using a local server, we can make it configurable. Want to pair on it?

trcywu: yeah sure

PR #84, opened 2020-07-29
first stab at kratos using their helm chart

alsuren:
TODO:

- [x] use kustomize to make the configmaps like suggested at https://redbadger.slack.com/archives/C013X35AGQ0/p1596101671003800
- [x] try to think of a way to get this working for both local development (yarn dev) and when you're actually running in the cluster. I think that https://www.npmjs.com/package/https-localhost might work for local development. I will try that first.
- [x] ~try to specify "/" in a bunch of places, without specifying the domain name~ no dice.
- [x] why is the docker image by default called `oryd/kratos:v0.4.3-alpha.1-sqlite`? Is there a slimmer version with only postgresql support?
- [x] add ticket about finding an email provider
- [x] scale kratos to 2 instances so we don't get an outage when one of them fails their health check.

Things to come in a separate PR:

- stop using --dev, and use https://www.npmjs.com/package/https-localhost instead

To test:
- [ ] `terraform apply`
- [ ] `edit infrastructure/dev-overlay-variables.json`
- [ ] `install-argo-cd.sh dev-$FNHSNAME`
- [ ] `sudo bash -c "while true; do kubefwd services -n argocd -n kratos -n hello-world; echo restarting in 1s; sleep 1; done"` - to forward the appropriate ports to make login work.
- [ ] `kubectl exec kratos-76fdc89bf4-hzvgn -n kratos -c kratos -it sh` then `kratos -c /etc/config/config.yaml migrate sql -e --yes` (FIXME: I think there is an auto-migrate option for kratos)
- [ ] `(cd ~/src/kratos-selfservice-ui-node/ &&  KRATOS_PUBLIC_URL=http://kratos-public.kratos/  KRATOS_ADMIN_URL=http://kratos-admin.kratos/  PORT=4455  SECURITY_MODE=cookie   npm start)` and create a new user (then tear their example ui down because it uses the same port as `frontend`
- [ ] `(cd frontend/ && yarn && yarn dev)`


infrastructure/kubernetes/argocd/apps/production/kratos.yaml
```diff
@@ -0,0 +1,18 @@
...
+  name: kratos
+spec:
+  destination:
+    namespace: kratos
+    server: https://kubernetes.default.svc
+  project: default
+  source:
+    # There is no difference between dev and prod other than secrets,
+    # and secrets are managed by terraform.
+    path: ./kratos/base
```
frigus02, line 13: Since we have no different in environments, I might get rid of the `base` folder and put files directly into the `./kratos` folder. But I can see the benefit of having the same layout everywhere. So maybe we leave it like this 🤔

infrastructure/kubernetes/argocd/apps/dev-template/kratos.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: argoproj.io/v1alpha1
+kind: Application
+metadata:
+  name: ingress
```
frigus02, line 4: I think that should be `kratos`
```suggestion
  name: kratos
```

infrastructure/kubernetes/kratos/base/kratos.yaml
```diff
@@ -0,0 +1,277 @@
...
+              path: /health/ready
+              port: http-admin
+          resources:
+            null
+---
+# Source: kratos/templates/tests/test-connection.yaml
+apiVersion: v1
+kind: Pod
+metadata:
+  name: "kratos-test-connection"
```
frigus02, line 262: I wonder if we need that pod or if this is only for helm hooks.

infrastructure/kubernetes/kratos/base/kratos.yaml
```diff
@@ -0,0 +1,277 @@
...
+      annotations:
+    spec:
+      volumes:
+        -
+          name: kratos-config-volume
+          configMap:
+            name: kratos-config
+      containers:
+        - name: kratos
+          image: "oryd/kratos:v0.4.3-alpha.1-sqlite"
```
frigus02, line 204: SQLite?

infrastructure/kubernetes/kratos/base/kratos.yaml
```diff
@@ -0,0 +1,277 @@
...
+        verification:
+          after:
+            default_browser_return_url: http://127.0.0.1:4455/
+          enabled: true
+          ui_url: http://127.0.0.1:4455/verify
+      strategies:
+        password:
+          enabled: true
+      whitelisted_return_urls:
+      - http://127.0.0.1:4455
```
frigus02, line 76: This is the local frontend URL, isn't it? I assume this means we need to patch this somehow for the different environments.

infrastructure/kubernetes/kratos/base/mailslurper.yaml
```diff
@@ -0,0 +1,47 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: mailslurper
```
frigus02, line 4: That's nice for development. I guess we need a ticket to attach this to a proper email service.

infrastructure/kubernetes/kratos/base/kratos.yaml
```diff
@@ -0,0 +1,277 @@
+# This file is generated by running `helm template -f ~/tmp/values.yaml ory/kratos --namespace kratos`
```
frigus02, line 1: Since we're editing it manually anyway and the file is not very complex, I wonder if we should go all in and basically write the file ourselves.

infrastructure/kubernetes/argocd/apps/dev-template/kratos.yaml
alsuren: oh oops. That's why it tried to deploy master when I deployed it.

infrastructure/kubernetes/kratos/base/kratos.yaml
alsuren: I was hoping that I could specify "/" in a bunch of places, without specifying the domain name. This is the next thing to try after refactoring to use kustomize to build the config maps. If it doesn't work then I will have to do the dev-template dance :( .

infrastructure/kubernetes/kratos/base/kratos.yaml
alsuren: not sure why they've decided to put that in their docker image tag. I will go see if there are any other tags that make more sense.

infrastructure/kubernetes/kratos/base/kratos.yaml
alsuren: Sounds like there is already a readiness probe. Happy to kill this.

infrastructure/kubernetes/kratos/base/mailslurper.yaml
alsuren: Yes, and probably also one to work out how to style the emails. https://www.ory.sh/kratos/docs/concepts/email-sms#templates says it will happen in "a future version", which doesn't inspire confidence. There might be a way that we can proxy all kratos mail through a service that adds our styling or something.

infrastructure/kubernetes/kratos/base/kratos.yaml
```diff
@@ -0,0 +1,138 @@
...
+kind: Service
+metadata:
+  name: kratos-admin
+  namespace: kratos
+  labels:
+    app.kubernetes.io/name: kratos
+    helm.sh/chart: kratos-0.4.0
+    app.kubernetes.io/instance: kratos
+    app.kubernetes.io/version: "0.4.3-alpha.1"
+    app.kubernetes.io/managed-by: Helm
```
alsuren, line 15: how many of these labels can I get away with removing?

infrastructure/kubernetes/kratos/base/kratos.yaml
frigus02: I mean, it doesn't matter in the end. Just a bit confusing 😄

infrastructure/kubernetes/kratos/base/kratos.yaml
frigus02: Probably all of them. The deployment or podspec in the deployment needs at least one label. The rest doesn't really matter.

I suppose you could keep "app.kubernetes.io/name: kratos". Makes it easy to query all Kratos resources if we need to.

infrastructure/modules/platform/main.tf
```diff
@@ -24,6 +24,15 @@ resource "azurerm_subnet" "cluster_nodes" {
   address_prefixes     = ["10.240.0.0/16"]
 }
 
+# FIXME: this is completely overkill permissions (owner of the terraform-created resource group)
+# Can we pin this down to just the cluster_outbound ip address?
+resource "azurerm_role_assignment" "managed_identity_operator_cluster" {
+  scope                            = azurerm_resource_group.platform.id
+  role_definition_name             = "Owner"
```
frigus02, line 31: The cluster wants to assign the IP to it's network. It needs the permission "Microsoft.Network/publicIPAddresses/join/action" for that. According to this documentation the role "Network Operator" should be enough.

https://docs.microsoft.com/en-us/azure/virtual-network/virtual-network-public-ip-address

alsuren:
```
Terraform will perform the following actions:

  # module.platform.azurerm_role_assignment.managed_identity_operator_cluster must be replaced
-/+ resource "azurerm_role_assignment" "managed_identity_operator_cluster" {
      ~ id                               = "/subscriptions/4a4be66c-9000-4906-8253-6a73f09f418d/resourceGroups/platform-dev-david/providers/Microsoft.Authorization/roleAssignments/466b6ee9-0d8d-0d93-9993-bf43ac4d153b" -> (known after apply)
      ~ name                             = "466b6ee9-0d8d-0d93-9993-bf43ac4d153b" -> (known after apply)
        principal_id                     = "5509bd74-6f90-4e90-b052-4042532743f6"
      ~ principal_type                   = "ServicePrincipal" -> (known after apply)
      ~ role_definition_id               = "/subscriptions/4a4be66c-9000-4906-8253-6a73f09f418d/providers/Microsoft.Authorization/roleDefinitions/8e3af657-a8ff-443c-a75c-2fe8c4bcb635" -> (known after apply)
      ~ role_definition_name             = "Owner" -> "Network Operator" # forces replacement
        scope                            = "/subscriptions/4a4be66c-9000-4906-8253-6a73f09f418d/resourceGroups/platform-dev-david"
        skip_service_principal_aad_check = true
    }
```
```
Error: Error loading Role Definition List: could not find role 'Network Operator'

  on ../../modules/platform/main.tf line 29, in resource "azurerm_role_assignment" "managed_identity_operator_cluster":
  29: resource "azurerm_role_assignment" "managed_identity_operator_cluster" {
```
no dice.

frigus02: Wow, that's because I can't read. In the link it clearly says the role is called **Network Contributor**. I don't know how I got to "Operator". Sorry.

> To perform tasks on public IP addresses, your account must be assigned to the network contributor role or to a custom role that is assigned the appropriate actions listed in the following table:

infrastructure/kubernetes/kratos/base/mailslurper.yaml
alsuren: crated ticket here: https://airtable.com/tblgtzV3sSwquTkIa/viwKLAjXIngcG7U7E/recVU3BGN3UCVbe1w?blocks=hide

(Ben said he's fine with un-styled emails for private beta)

infrastructure/kubernetes/kratos/dev-template/config.yaml
```diff
@@ -0,0 +1,68 @@
...
+      ui_url: $KRATOS_WEB_ROOT/settings
+      privileged_session_max_age: 15m
+
+    recovery:
+      enabled: true
+      ui_url: $KRATOS_WEB_ROOT/recovery
+
+    verification:
+      enabled: true
+      ui_url: $KRATOS_WEB_ROOT/verify
```
frigus02, line 32: I wonder if those URL should be `/auth/verify`. I could imagine that we want to use something like `/settings` for some non-auth settings.

But we're not using those, yet. So let's think about this later.

frigus02: Should we disable all flows that we're not currently using?

frigus02: Also in production? 🙂

PR #89, opened 2020-08-03
Ensure all developers have access to key vaults

frigus02:
Before this, only the person applying Terraform will get permissions to analytics secrets. If another person then tries to apply Terraform, they will get the following error:

    Error: Error making Read request on Azure KeyVault Secret synapse-password: keyvault.BaseClient#GetSecret: Failure responding to request: StatusCode=403 -- Original Error: autorest/azure: Service returned an error. Status=403 Code="Forbidden" Message="The user, group or application 'appid=XXX;oid=XXX;numgroups=1;iss=https://sts.windows.net/XXX/' does not have secrets get permission on key vault 'fnhs-analytics-XXX;location=westeurope'. For help resolving this issue, please see https://go.microsoft.com/fwlink/?linkid=2125287" InnerError={"code":"AccessDenied"}

This gives the FutureNHS Developers AAD group permissions instead, which means all core developers are able to apply Terraform.


frigus02:
Unfortunately:

> Error: Error initializing PostgreSQL client: error detecting capabilities: error PostgreSQL version: pq: no pg_hba.conf entry for host "82.47.239.183", user "psqladminun", database "postgres", SSL on

I guess we can work with this for now. Just means whoever applies terraform needs to apply the platform module first.

PR #90, opened 2020-08-03
Workspace Service

matt-paul:
Screenshot 2020-08-04 at 17 45 06

[Ticket](https://airtable.com/tblgtzV3sSwquTkIa/viw8ZHoNqvSvl62lt/recMsd5a69XusQz6p?blocks=hide)

as a frontend developer
I want to have a backend service for listing workspaces
so that I can query it when rendering webpages

A read-only list of names is probably fine for a first step. If you feel like letting people add their own workspace names, then that's fine.

## Acceptance Criteria
- [x] Workspace database added to terraform.
- [x] Create a graphql service that intereacts with the database via GraphQL playground
  - [x] Create
  - [x] Read
  - [x] Update
  - [x] Delete
- [x] Works locally
- [x] Create temp db via service container for sqlx to verify within CI
- [x] Works in dev cluster via port forwarding (manual apply of migrations).
- [ ] Migrations are run on dev database
- [ ] Production Cluster

## Testing information
1) Apply terraform changes to your cluster.
2) Connect Tableplus to workspace database.
3) Create schema:
```
CREATE TABLE IF NOT EXISTS workspace (
    id uuid DEFAULT uuid_generate_v4 (),
    title TEXT NOT NULL,
    PRIMARY KEY (id)
);
```

4)Load up the Graphql playground by using kubectl port-forward. 
`kubectl port-forward -n workspace-service deployments/workspace-service 3030:3030`

- find_all
```
{
  workspaces {
    id,
    title
  }
} 
```

- find_one
```
{
  workspace (id: "ID_OF_AN_ENTRY"){
    id,
    title
  }
} 
```

- create
```
mutation {
  createWorkspace(workspace: {title: "Covid"}) {
    id,
    title
  }
}
```

- update
```
mutation {
  updateWorkspace(id: "ID_OF_AN_ENTRY", workspace: { title: "Covida"}) {
    id,
    title
  }
}
```

- delete
```
mutation {
  deleteWorkspace(id: "ID_OF_AN_ENTRY") {
    id,
    title
  }
}
```


cobwebfx:
test comment

infrastructure/modules/databases/main.tf
```diff
@@ -1,53 +1,64 @@
 # This requires the postgresql and kubernetes providers to be set up properly.
 # Run `terraform apply -target module.platform` to set up their requirements
 # before attempting to apply this module.
+locals {
+  databases = [
+    "kratos-db",
+    "workspace-db",
```
matt-paul, line 7: Annoyingly, applying to the cluster with simple 'kratos' and 'workspace' names caused terraform to error as kratos namespace already existed.  Perhaps my terraform-fu was not good enough, will come back to this and simplify if possible to remove appended '-db'.

infrastructure/modules/databases/main.tf
```diff
@@ -18,48 +25,57 @@ resource "random_password" "kratos_secrets_cookie" {
...
   metadata {
-    name      = "kratos"
-    namespace = "kratos"
+    name      = "${each.value}-db-creds"
+    namespace = each.value
   }
   data = {
     secretsDefault = random_password.kratos_secrets_default.result
     secretsCookie  = random_password.kratos_secrets_cookie.result
     dsn = "postgres://${
```
matt-paul, line 69: TODO: This was a mistake to try to iterate this, do this individually as most of these are kratos only

workspace-service/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+
+RUN useradd svc
+
+# Build real binaries now, as late as possible
+FROM build-context AS build
+
+COPY ./src ./src
+COPY sql ./sql
+
+RUN cargo clippy --release -- -D warnings && cargo test --release && cargo build --release
```
matt-paul, line 37: With autoformatting turned on in VSCode, the split line version of this (as appears in other dockerfiles) gets incorrectly formatted on save (and therefore fails CI.  This is why this is now a one liner.  If someone could try the same in their VSCode it would be helpful to know whether i have some sort of conflict in my linting settings.

.github/workflows/workspace-service-branch.yaml
```diff
@@ -0,0 +1,121 @@
...
+    runs-on: ubuntu-latest
+    services:
+      # Label used to access the service container
+      postgres:
+        # Docker Hub image
+        image: postgres
+        # Provide the password for postgres
+        env:
+          POSTGRES_PASSWORD: postgres
+        # Set health checks to wait until postgres has started
```
frigus02, line 21: We can probably remove some of the comments.

.github/workflows/workspace-service-branch.yaml
```diff
@@ -0,0 +1,121 @@
...
+      - name: Connect to PostgreSQL
+        working-directory: futurenhs-platform/workspace-service
+        run: |
+          set -e
+          cargo install --version=0.1.0-beta.1 sqlx-cli --no-default-features --features postgres
+          export DATABASE_URL=postgres://postgres:postgres@localhost:5432
+          sqlx migrate run
+        env:
+          POSTGRES_HOST: postgres
+          POSTGRES_PORT: 5432
```
frigus02, line 52: These are not used if I see this correctly. So we might as well remove the env vars.

.github/workflows/workspace-service-branch.yaml
```diff
@@ -0,0 +1,121 @@
...
+      - name: "Login to docker"
+        uses: azure/docker-login@v1
+        with:
+          login-server: ${{ secrets.REGISTRY_LOGIN_SERVER }}
+          username: ${{ secrets.REGISTRY_USERNAME }}
+          password: ${{ secrets.REGISTRY_PASSWORD }}
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
+      - name: Connect to PostgreSQL
```
frigus02, line 43: Maybe we can call the step something like "Run DB migrations" or "Initialize test database"

.github/workflows/workspace-service-master.yaml
```diff
@@ -0,0 +1,78 @@
...
+        uses: actions/checkout@v2
+        with:
+          path: futurenhs-platform
+      - name: "Login to docker"
+        uses: azure/docker-login@v1
+        with:
+          login-server: ${{ secrets.REGISTRY_LOGIN_SERVER }}
+          username: ${{ secrets.REGISTRY_USERNAME }}
+          password: ${{ secrets.REGISTRY_PASSWORD }}
+      - name: "Build & Push image"
```
frigus02, line 25: This will not work, yet, because of the missing DB. (I know you know that. Just leaving it here as a todo 🙂)

README.md
```diff
@@ -4,6 +4,10 @@
 
 NOTE: You can find information on setting up the cluster and relevent tooling [here](https://github.com/FutureNHS/futurenhs-platform/blob/master/infrastructure/README.md).
 
+## Connecting to databases in your cluster
+
+[tableplus](https://www.tableplus.io/download) might be useful.
```
frigus02, line 9: I'm always on the fence about suggesting tools. But I guess this is fine. This should be a level 3 heading, though 🔍 🙂

infrastructure/kubernetes/ingress/production/ingress-workspace-service.yaml
```diff
@@ -0,0 +1,22 @@
+apiVersion: extensions/v1beta1
```
frigus02, line 1: The Ingress is not needed. We don't want the service to be publicly available. Because then we would need to make an authentication check.

Instead we route everything through the frontend for now, which does the authentication checks.

infrastructure/modules/databases/main.tf
```diff
@@ -18,25 +25,33 @@ resource "random_password" "kratos_secrets_cookie" {
   upper   = true
 }
 
-resource "postgresql_role" "kratos" {
-  name     = "kratos_user"
+resource "postgresql_role" "service" {
+  for_each = toset(local.databases)
+  name     = "${each.value}_user"
```
frigus02, line 30: The resulting username for the workspace service is a bit ugly in my opinion (`workspace-service_user`). I'm in favor of removing the `_user` suffix, but that would be a change to kratos. Should we instead use underscores in the DB names, i.e. `workspace_service`?

infrastructure/modules/databases/main.tf
```diff
@@ -18,25 +25,33 @@ resource "random_password" "kratos_secrets_cookie" {
...
   allow_connections = true
 }
 
-resource "kubernetes_namespace" "kratos" {
+resource "postgresql_extension" "uuid-ossp" {
+  name     = "uuid-ossp"
+  database = "workspace-service"
+}
+
+resource "kubernetes_namespace" "db" {
```
frigus02, line 51: Let's name the resource `service` as well.
```suggestion
resource "kubernetes_namespace" "service" {
```

workspace-service/Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+
+RUN rustup component add rustfmt clippy
+
+WORKDIR /usr/src/workspace-service
+
+COPY Cargo.toml .
+COPY Cargo.lock .
+
+# IP Address tells docker to talk to host
+ENV DATABASE_URL=postgres://postgres:postgres@172.17.0.1:5432
```
frigus02, line 12: We probably want to make this configurable, so that it works locally. Maybe we a build arg.

```
ARG DATABASE_URL=postgres://postgres:postgres@172.17.0.1:5432 
ENV DATABASE_URL=$DATABASE_URL
```

Then we can pass in our locally valid URL.

```
docker build --build-arg DATABASE_URL=my-local-url
```

workspace-service/Dockerfile
frigus02: I noticed that as well. I'm not sure what VS Code is doing there.

workspace-service/manifests/base/namespace.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
+kind: Namespace
```
frigus02, line 2: The namespace should not be needed because it's created in terraform.

workspace-service/manifests/dev-template/deployment.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: apps/v1
+kind: Deployment
```
frigus02, line 2: This doesn't seem to patch anything. Should we remove this for now?

workspace-service/src/db.rs
```diff
@@ -0,0 +1,49 @@
...
+
+impl Workspace {
+    pub async fn create(title: String, pool: &PgPool) -> Result<Workspace> {
+        let workspace = sqlx::query_file_as!(Workspace, "sql/create.sql", title)
+            .fetch_one(pool)
+            .await?;
+
+        Ok(workspace)
+    }
+    pub async fn find_all(pool: &PgPool) -> Result<Vec<Workspace>> {
```
frigus02, line 18: Empty line missing between the create and find_all methods 🔍

workspace-service/src/graphql.rs
```diff
@@ -0,0 +1,124 @@
...
+
+pub struct QueryRoot;
+
+#[Object]
+impl QueryRoot {
+    #[field(description = "Get all Workspaces")]
+    async fn workspaces(&self, context: &Context<'_>) -> FieldResult<Vec<Workspace>> {
+        let pool = context.data()?;
+        let workspaces = db::Workspace::find_all(pool).await?;
+        Ok(workspaces.iter().cloned().map(Into::into).collect())
```
frigus02, line 60: I wonder if we need to clone here. Could we use `workspaces.into_iter().map(Into::into).collect()` instead?

workspace-service/src/graphql.rs
```diff
@@ -0,0 +1,124 @@
...
+        workspace: UpdateWorkspace,
+    ) -> FieldResult<Workspace> {
+        let pool = context.data()?;
+        let workspace =
+            db::Workspace::update(Uuid::parse_str(id.as_str())?, workspace.title, pool).await?;
+
+        Ok(workspace.into())
+    }
+
+    #[field(description = "Delete workspace(returns updated workspace")]
```
frigus02, line 101:
```suggestion
    #[field(description = "Delete workspace (returns deleted workspace")]
```

workspace-service/src/graphql.rs
```diff
@@ -0,0 +1,124 @@
...
+    async_graphql_tide::graphql(req, schema, |query_builder| query_builder).await
+}
+
+pub async fn handle_graphiql(_: Request<State>) -> tide::Result {
+    let mut response = Response::new(StatusCode::Ok);
+    response.set_body(Body::from_string(playground_source(
+        GraphQLPlaygroundConfig::new("/graphql"),
+    )));
+
+    response.set_content_type(mime::HTML);
```
frigus02, line 121: Tide has a builder for this now. So we could write this as

```rust
let response = Response::builder(StatusCode::Ok)
    .body(playground_source(GraphQLPlaygroundConfig::new("/graphql")))
    .content_type(mime::HTML)
    .build();
```

workspace-service/src/main.rs
```diff
@@ -0,0 +1,31 @@
...
+async fn main() -> Result<()> {
+    dotenv().ok();
+    let instrumentation_key =
+        env::var("INSTRUMENTATION_KEY").expect("env var INSTRUMENTATION_KEY should exist");
+    let exporter = opentelemetry_application_insights::Exporter::new(instrumentation_key);
+    let batch_exporter =
+        BatchSpanProcessor::builder(exporter, tokio::spawn, tokio::time::interval).build();
+    let provider = sdk::Provider::builder()
+        .with_batch_exporter(batch_exporter)
+        .build();
```
frigus02, line 18: I wonder if we should allow this to run with an instrumentation key locally. If no instrumentation is present, we could instead build an exporter, which logs to stdout:

```rust
let exporter = opentelemetry::exporter::trace::stdout::Builder::default().init();
let provider = opentelemetry::sdk::Provider::builder()
    .with_simple_exporter(exporter)
    .build();
```

workspace-service/src/graphql.rs
matt-paul: Cool, so that would be a more idiomatic way of writing Rust?

frigus02: In this example it doesn't add much. The only thing is that the `response` variable doesn't have to be mutable. But the function is so small that it doesn't really matter. You can easily see how the variable is used.

Builders are quite common in Rust, though. They can let you write code with optional values quite nicely and safely.

You could possibly argue that the builder is slightly more readable here. But I think this might be subjective.

matt-paul: Makes sense, thanks.

workspace-service/src/graphql.rs
matt-paul: That works.  The advantage being that we use less memory by referencing the the workspaces instead?

frigus02: Yeah. We already have the workspaces in memory and we're not using them anymore after this. So there is no need to create another copy of them.

.github/workflows/workspace-service-master.yaml
alsuren: If we don't have any integration tests that use the DB, we should be able to avoid the db dependency in CI by checking it into the repo. I would expect it to look something like this: https://github.com/redbadger/feature-targeting/blob/master/examples/todomvc/api/sqlx-data.json

frigus02: I thought that as well. But we can't check if this file is up to date without connecting to the database.

frigus02: The documentation is here: https://github.com/launchbadge/sqlx/tree/master/sqlx-cli#enable-building-in-offline-mode-with-query

Essentially, we would like to run `sqlx prepare --check`. But that requires a DB connection.

We could say that we can skip this for master builds. And we rely on the proper check in branch builds, though. That would make master builds quicker, which is nice.

workspace-service/src/main.rs
matt-paul: This comment confuses me i'm afraid.  Do you mean ''run *without* an instrumentation key locally"?

frigus02: Well, at the moment the application crashes if you don't have an INSTRUMENTATION_KEY environment variable, doesn't it? That's a bit annoying.

You could do something like:

```rust
if let Some(instrumentation_key) = env::var("INSTRUMENTATION_KEY") {
  // create application insights exporter
} else {
  // create stdout exporter
}
```

matt-paul: Gotcha, just wanted to double check

.github/workflows/workspace-service-master.yaml
matt-paul: @alsuren Just chatting with Jan, would be great to chat about this point tomorrow after standup in case there is anything we have missed, and then I will implement result of conversation in a fresh PR

frigus02: We talked this through.

- In this PR we use the service container for master

- In another PR we use the SQLx JSON file to make the Docker build not reliant on a DB

- Use actions/cache for sqlx cli

workspace-service/Dockerfile
frigus02: Not needed. We will make the Docker build not reliant on a DB using the json file.

.github/workflows/workspace-service-master.yaml
matt-paul: Result of discussion: 2 further PR's to come
1) Cache sqlx build so that it CI doesn't take so long
2) Decouple docker build and service container by using SQLX json file approach.  The service container check will remain.

infrastructure/kubernetes/ingress/dev-template/kustomization.yaml
```diff
@@ -24,5 +24,12 @@ patchesJSON6902:
       kind: Ingress
       name: ingress
       namespace: kratos
+  - path: patch-ingress.yaml
+    target:
+      group: extensions
+      version: v1beta1
+      kind: Ingress
+      name: ingress
+      namespace: workspace-service
```
frigus02, line 33: We don't need this ingress patch anymore. And we can also remove the hello-world one here.

workspace-service/manifests/base/kustomization.yaml
```diff
@@ -0,0 +1,5 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+  - namespace.yaml
```
frigus02, line 4: The namespace.yaml does not exist anymore, so this reference can go.

workspace-service/manifests/production/deployment.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: apps/v1
+kind: Deployment
```
frigus02, line 2: Similar to the dev one, this doesn't patch anything. Let's remove it for now.

.github/workflows/workspace-service-master.yaml
```diff
@@ -0,0 +1,78 @@
...
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - workspace-service/**
+      - .github/workflows/workspace-service-master.yaml
+
+jobs:
+  ci:
```
frigus02, line 12: We need to postgres service container for this build as well, because it will fail otherwise.

workspace-service/Dockerfile
```diff
@@ -0,0 +1,57 @@
+# Setup rust build environment
+FROM rust:1.44.1 AS build-context
```
alsuren, line 2: https://github.com/FutureNHS/futurenhs-platform/pull/90/checks?check_run_id=981178610#step:7:1034 (` error[E0658]: procedural macros cannot be expanded to expressions`) suggests that we should be using rust:1.45.2

PR #91, opened 2020-08-03
make yarn dev use https

alsuren:
Outstanding issues:

- [ ] Once we remove the --dev option, there is no way to bootstrap users into the system. This is because the example app doesn't use https, and so won't be able to use any of the cookies that kratos sends.
  - this is part of https://airtable.com/tblgtzV3sSwquTkIa/viwKLAjXIngcG7U7E/recLaXEtzIwsLMJXb?blocks=hide

### DIFF
--- a/frontend/.gitignore
+++ b/frontend/.gitignore
@@ -1,5 +1,8 @@
 # See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
 
+# SSL certificates
+localhost*.pem
+
 # dependencies
 /node_modules
 /.pnp
--- a/frontend/README.md
+++ b/frontend/README.md
@@ -6,10 +6,18 @@ This is a [Next.js](https://nextjs.org/) project bootstrapped with [`create-next
 
 Frontend development server relies on some services that are available in the development cluster. Specifically, it calls out to the hello-world service when doing server-side rendering, and also proxies `/hello/*` to `http://hello-world.hello-world/hello/*` like how our nginx ingress does in production.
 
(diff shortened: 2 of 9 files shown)


frigus02:
I was interested in how this would work with [Caddy](https://caddyserver.com/). I think it's actually pretty easy. 

- You need to install it with `brew install caddy` (and `brew install nss` for Firefox support).
- Then make a config file called `Caddyfile` in the frontend folder

  ```
  localhost:2015

  route /.ory/kratos/public/* {
  	uri strip_prefix /.ory/kratos/public
  	reverse_proxy http://kratos-public.kratos
  }

  reverse_proxy 127.0.0.1:4455
  ```
- Then start it with either `caddy start` (runs in background) or `caddy run` (runs in foreground)

That replaces the server.js file completely, I think. And does automatic TLS with generated certificates installed in the local cert store.

There is however a weird bug. I need to run `caddy run` as `sudo` for some reason because it tries to bind to port 80.

If we need to configure more stuff, the JavaScript solution could be nicer, though. It's not another config language. It also runs only a single process.

frigus02:
Talked with @alsuren  earlier.

We actually need `--dev` in order to access the user registration flow at the moment. Without that we're not able to register users.

This changes once we implement the password reset flow properly through the frontend. Me may park this PR until this is done.

alsuren:
I'm assuming that this is no longer useful now we've switched to azure's login solution, so it would just add complexity that we don't need. Closing.

PR #96, opened 2020-08-05
Event sending infrastructure

frigus02:
[Ticket](https://airtable.com/tblgtzV3sSwquTkIa/viw8ZHoNqvSvl62lt/recJ2EpMnCMF62C4P)

This is the first part of the ticket.
- It covers infrastructure plus an example of sending events.
- Todo is the pipeline to push the event to the Analytics storage.

In this PR you will find:
- Documentation of the proposed event flow. This shows how it's used by other applications and for analytics.
- Infrastructure to create an Event Grid Topic and analytics Event Hub. This enables us to send events. And it proves that this whole thing will work.
- The frontend sends an example login attempt event. Locally this only results in a console.log unless the corresponding environment variables have been set.
- The frontend has a health endpoint now, which checks if the environment variables exist. This is to ensure that in a cluster, it will always send proper events.

More info:
- For Rust services, we can publish events without an SDK by calling their API manually: https://docs.microsoft.com/en-us/rest/api/eventgrid/dataplane/publishevents/publishevents
- The frontend and hello-world namespace is now defined in Terraform. To apply this to dev cluster you need to:
  - Tell Argo CD about the changes (this will delete the frontend namespace and result in a sync error)
  - Apply Terraform
  - Tell Argo CD to sync again. This will deploy the frontend again

## Pre-review checklist:

- [x] Tests

- [x] Documentation

## Test:

- [x] Tested locally

- [x] Tested on dev cluster


frontend/pages/api/health.ts
```diff
@@ -0,0 +1,14 @@
+import { NextApiRequest, NextApiResponse } from "next";
+
+export default (_: NextApiRequest, res: NextApiResponse) => {
+  if (
+    !process.env.EVENTGRID_TOPIC_ENDPOINT ||
+    !process.env.EVENTGRID_TOPIC_KEY
+  ) {
+    res
+      .status(500)
+      .json({ status: "Error", message: "Missing environment variables" });
```
alsuren, line 10: how would I debug which environment variable is missing if I saw this error message?

infrastructure/dev-overlay-variables.json
```diff
@@ -22,7 +22,7 @@
   "jan": {
     "BRANCH": "master",
     "KRATOS_WEB_ROOT": "https://fnhs-dev-jan.westeurope.cloudapp.azure.com",
-    "INSTRUMENTATION_KEY": "14fe00c5-310f-4e84-aa83-5db7bd5b2797"
+    "INSTRUMENTATION_KEY": "53bd5e02-2b32-47f5-989d-09f7b16a396f"
```
alsuren, line 25: ... we should really rip this out and provide it as a secret from terraform. (for another PR)

infrastructure/modules/databases/main.tf
```diff
@@ -63,3 +64,28 @@ resource "kubernetes_secret" "kratos_db_creds" {
     }"
   }
 }
+
+data "azurerm_eventgrid_topic" "platform" {
```
alsuren, line 68: looks like databases was the wrong name for this module after all.

infrastructure/modules/databases/main.tf
```diff
@@ -63,3 +64,28 @@ resource "kubernetes_secret" "kratos_db_creds" {
...
+  name                = var.eventgrid_topic_name
```
alsuren, line 69: why does this need to be configurable?

... oh, so that terraform will realise that there is a dependency between the two modules or something?

infrastructure/modules/databases/main.tf
alsuren, line 68: I thought we agreed we weren't going to use `data` in the databases module, and we would use variables instead.

infrastructure/modules/databases/main.tf
```diff
@@ -63,3 +64,28 @@ resource "kubernetes_secret" "kratos_db_creds" {
...
+  resource_group_name = var.resource_group_name
+}
+
+resource "kubernetes_namespace" "frontend" {
```
alsuren, line 73: are we doing hello-world as well, while we're here?

infrastructure/dev-overlay-variables.json
frigus02: Agree. I want to put this in a configmap for all services that need it.

frontend/lib/events.ts
```diff
@@ -0,0 +1,30 @@
+import { v4 as uuid } from "uuid";
+import EventGridClient from "azure-eventgrid";
+import { TopicCredentials } from "ms-rest-azure";
+
+export interface Event {
+  subject: string;
+  eventType: string;
+  data: any;
+  dataVersion: string;
```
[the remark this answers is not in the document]
frigus02: Yeah, we definitely need a proper schema for this. I'm not sure how to do that. It would be amazing to generate both TypeScript and Rust models from some kind of schema config.

Maybe we can use OpenAPI for this?

frontend/lib/test-helpers/environment.ts
```diff
@@ -0,0 +1,11 @@
+export const withEnvVars = (vars: { [name: string]: string }) => {
+  beforeAll(() => {
+    Object.assign(process.env, vars);
+  });
+
+  afterAll(() => {
+    for (const key of Object.keys(vars)) {
+      delete process.env[key];
```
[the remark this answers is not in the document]
frigus02: Good idea 👍

infrastructure/modules/databases/main.tf
frigus02: Yeah, why not. Let's do hello-world now. Than we don't have to fiddle around with this again later on.

frontend/lib/events.ts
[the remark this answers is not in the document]
frigus02: I generated TypeScript and Rust models here: https://github.com/FutureNHS/futurenhs-platform/pull/99

It will always generate an HTTP client as well. But I guess we could write a small shell script, that deletes the client.

The models look alright to me.

However it's not trivial to define this type of model in OpenAPI. I did not find a nice way of defining the common properties like subject, eventType and dataVersion only once. But also I haven't tried everything, yet.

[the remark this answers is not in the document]
frigus02: I'm going to change this PR to have strong types for this one particular case. Just to that it's clear it's not going to stay that way.

infrastructure/modules/databases/main.tf
frigus02: Yeah. I see 2 options:

- Rename module to something like "app_config"
- Make separate modules for the different things. E.g. 1 database module, 1 kubernetes module or maybe 1 database module and 1 event module

What do you prefer?

infrastructure/dev-overlay-variables.json
frigus02: I can make a separate PR to do that.

infrastructure/modules/databases/main.tf
frigus02: The name changes per environment. And yeah, also because of the dependency.

infrastructure/modules/databases/main.tf
frigus02: I was hesitant to add so many different variables. But actually, it turns out it's 2 variables either way. Let me change that.

infrastructure/dev-overlay-variables.json
frigus02: Okay. It's a draft for now, because it depends on the namespace being created here.
https://github.com/FutureNHS/futurenhs-platform/pull/101

PR #99, opened 2020-08-06
Event model generation using OpenAPI

frigus02:
- Are those event complete? Do they contain enough information to reconstruct the state of workspace if you have all the event?
- How do we deal with permissions or data deletion of fields in those events? Do we need to encrypt fields inside the events?


event-models/schema.yaml
```diff
@@ -0,0 +1,55 @@
...
+  version: "1.0.0"
+  title: "FutureNHS Events"
+paths: {}
+components:
+  schemas:
+    Event:
+      oneOf:
+        - $ref: "#/components/schemas/LoginFailed"
+        - $ref: "#/components/schemas/LoginSuccess"
+      discriminator:
```
alsuren, line 12: I searched for openapi discriminator and this came up: `https://github.com/OAI/OpenAPI-Specification/issues/2143` ("Deprecate discriminator?". Not a link because I don't want our PR to show up on their issue). It might be worth looking into whether there are json schema type generators for the appropriate languages, so that we can skip generating a bunch of http interfaces that we don't need.

alsuren: Oh hey:https://json-schema.org/implementations#generators-from-schemas points to https://app.quicktype.io/# which is written in typescript and can output types for typescript and rust and go and a whole bunch of other languages if we need.

frigus02: That sounds amazing!!

frigus02: Unfortunately QuickType can't deal with unions at the moment: https://github.com/quicktype/quicktype/issues/1266

frigus02: Having 2 properties (`eventType` and `dataVersion`) define the data model is quite annoying.

TypeScript seems to be able to deal with this. So I guess for TypeScript I would expect models like rougly like this: [example in playground](https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgDIHsDmmIBMCSIAogG4TjIDeAUMnchGeACoCeADhAFzIBEANlhy5QvANy16uOGDgA1aAGdg6ED14kAjOMl1psnjXrHkAV0XQeisFFCYJxgL7Vn1UJFiIUGbHkKlyMDkAJipdBiYwNk51QV8REB1jfXklFTU+EmCkqRk4Q3DjYFwrGzsHEzoIAFs4YH5S2xB7cOdXd2h4JDQsUAAxOv48AIojekZA6O4+ONB4erwcvTyFKGVVdS0l5BSCyrpqiEVFOBxG8taXamowDhQAeVMoEbBkAF4e+P9I5AAfT+E30CIT+n36g2GkQk1BgphACDA6WQ7Cg6CQxwAFBNwDxHs9IgBKMJFGDILGRAB02Kid3ebw+AiEeFEyAAZKyIoEKSlVusQHSGVsiWNKghVIp0EMKXFyVyUhTzNACRU6G0gA)

In Rust this comes close to an [adjacently tagged enum in Serde](https://serde.rs/enum-representations.html#adjacently-tagged), but that supports only one tag and we need two.

frigus02:
Going to close this, as the JSON Schema one (#104)  is more promising.

PR #103, opened 2020-08-10
make kratos send password reset emails

alsuren:
User password reset flow looks like this:

## go to /auth/recovery and type in your email address
Screenshot 2020-08-11 at 11 36 11

## get told to check your emails
Screenshot 2020-08-11 at 11 36 33

## receive an email with a link
  * dev clusters will include the name of the cluster in the email address, and render as "via sendgrid.me"
  * prod will come from login.beta.future.nhs.uk and will not have the "via sendgrid.me" tag once some cname records have been recreated (see https://redbadger.slack.com/archives/C013X35AGQ0/p1597232925017500?thread_ts=1597156935.024500&cid=C013X35AGQ0)

Screenshot 2020-08-12 at 15 37 52

## click on the link
Screenshot 2020-08-11 at 11 43 11

## you get told to reset your password.
Screenshot 2020-08-11 at 11 45 22

- [ ] TODO: debug this error
Screenshot 2020-08-11 at 11 45 42

- [ ] TODO: if everything goes well, you get sent back to the password reset form. We can probably do better than this and redirect you to the homepage or somewhere?

- [ ] TODO: I think that the invalid certificate is the thing that is causing chrome to not suggest a password. We should fix this.

- [ ] The docker image is based on master from a few days ago (which has kratos pr https://github.com/ory/kratos/pull/634 merged)
and with the branch from pr https://github.com/ory/kratos/pull/617 manually merged in. The latter PR has now been merged, so we can just build an image from master now.

- [ ] We probably want to configure a kratos error page because this is not a great user experience when your csrf token is expired: `{"error":{"code":400,"status":"Bad Request","request":"7c552214cd4d3e3a3e3772815693be9d","reason":"CSRF token is missing or invalid.","message":"The request was malformed or contained invalid parameters"}}`

[Ticket](https://airtable.com/tblgtzV3sSwquTkIa/viwKLAjXIngcG7U7E/recVU3BGN3UCVbe1w?blocks=hide)

## Pre-review checklist:

- [ ] Tests

- [ ] Documentation

- [ ] Feature flag

## Testing information

* `sudo kubefwd services --context dev-david -n kratos`
* create a new user `./infrastructure/scripts/new-kratos-user.sh your.name+whatever@red-badger.com`. It is impossible to login at this point because they have no password.
* Trigger the password reset flow by going to 

## Test:

- [ ] Tested locally

- Platforms/Browsers:

  - [ ] Internet Explorer 11

  - [ ] Edge

- Accessibility:

  - [ ] Screen Reader

  - [ ] Keyboard Navigation

  - [ ] Magnification

  - [ ] Contrast

  - [ ] Text size 200%

  - [ ] Landscape (Mobile)


frontend/pages/auth/login.tsx
```diff
@@ -1,7 +1,8 @@
 import React from "react";
 import { GetServerSideProps } from "next";
-import { generateFields, FormConfig } from "../../lib/login";
+import { getLoginFields } from "../../lib/auth";
 import { sendEvent } from "../../lib/events";
+import { LoginRequestMethodConfig } from "@oryd/kratos-client";
 
 export const getServerSideProps: GetServerSideProps = async (context) => {
```
alsuren, line 7: hrm. This implicitly returns `any`. Is there a lint to catch this?

frigus02: I'm not sure what you mean. This function has an explicit return type:

```ts
export type GetServerSideProps<
  P extends { [key: string]: any } = { [key: string]: any },
  Q extends ParsedUrlQuery = ParsedUrlQuery
> = (context: GetServerSidePropsContext<Q>) => Promise<{ props: P }>
```

alsuren: Ah. I missed the `: GetServerSideProps` bit.

I suppose I would prefer to pin down P a bit harder, like I've done in recovery.tsx, but there's probably no way to enforce that with a linter.

frigus02: Oh right, but there is `any` in `P`. Unfortunately that's an explicit `any`. Not sure how to catch this 😕

.github/workflows/infrastructure-branch.yaml
```diff
@@ -66,11 +75,11 @@ jobs:
       - name: Deploy
         working-directory: futurenhs-deployments
         run: |
-          set -eux
+          set -euxo pipefail
 
           git add -A
           # TODO: can we make this commit message reflect the message at the tip of the platform repo?
-          git diff-index --quiet HEAD || git commit -am "CI: Update cert-manager and ingress"
+          git diff-index --quiet HEAD || git commit -am "infrastructure-branch: $COMMIT_MESSAGE"
```
alsuren, line 82: this causes commit messages that look like those on https://github.com/FutureNHS/futurenhs-deployments/commits/pr-send-emails/kratos

frontend/pages/auth/login.tsx
```diff
@@ -66,7 +68,7 @@ const Login = ({
                     name={name}
                     type={type}
                     required={required}
-                    defaultValue={value}
+                    defaultValue={(value as unknown) as string}
```
alsuren, line 71: - [ ] this feels like a bug in their openapi spec/codegen. I should submit it upstream.

frontend/pages/auth/settings.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const recovery = await getSettingsFields(request);
+  console.log(JSON.stringify(recovery, null, 4));
+  const formConfig = recovery.methods.password.config!;
+  const messages = recovery.messages?.map((msg) => msg.text ?? "") ?? null;
+  console.log(JSON.stringify(messages, null, 4));
+
+  // TODO: This is just an example event. We need to figure out the schema for custom events and change this to events we really need.
+  await sendEvent({
+    subject: "frontend",
+    eventType: "frontend.recovery.attempt",
```
alsuren, line 38:
```suggestion
    eventType: "frontend.settings.attempt",
```

frontend/pages/auth/recovery.tsx
```diff
@@ -0,0 +1,83 @@
...
+  if (!request && context.res) {
+    context.res.writeHead(302, {
+      Location: "/.ory/kratos/public/self-service/browser/flows/recovery",
+    });
+    context.res.end();
+    return { props: {} as any };
+  }
+  if (!request || Array.isArray(request)) {
+    return { props: {} as any };
+  }
```
frigus02, line 27: I wonder what we want to do in this case. Can this actually happen? Should we throw an error?

frontend/pages/auth/recovery.tsx
```diff
@@ -0,0 +1,83 @@
...
+  context
+): Promise<{ props: RecoveryProps }> => {
+  const request = context.query.request;
+
+  if (!request && context.res) {
+    context.res.writeHead(302, {
+      Location: "/.ory/kratos/public/self-service/browser/flows/recovery",
+    });
+    context.res.end();
+    return { props: {} as any };
```
frigus02, line 23: It might be nice if we could create a helper function for redirects. Basically I would like to have something descriptive that tells you why we return empty props. It's safe because our component will never see them.

frontend/pages/auth/recovery.tsx
```diff
@@ -0,0 +1,83 @@
...
+  }
+  if (!request || Array.isArray(request)) {
+    return { props: {} as any };
+  }
+
+  const recovery = await getRecoveryFields(request);
+  console.log(JSON.stringify(recovery, null, 4));
+  const formConfig = recovery.methods.link.config!;
+  const messages = recovery.messages?.map((msg) => msg.text ?? "") ?? null;
+  console.log(JSON.stringify(messages, null, 4));
```
frigus02, line 33: Do we want to remove the console.log statements, so they don't add noise to test outputs?

infrastructure/scripts/new-kratos-user.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+EMAIL="${1:?"Please specify your email as the first parameter, e.g. jane.doe@red-badger.com"}"
+
+TEMPLATE='{
+  "traits": {
+    "email": "EMAIL"
```
frigus02, line 9: I think in theory we could use `jq` to safely add variables to JSON. But that seems overkill for now.

```sh
  body=$(
    jq \
      --null-input \
      --arg email "$EMAIL" \
      '{
        traits: {
          email: $email
        }
      }'
  )
```

frontend/pages/auth/recovery.tsx
alsuren: TESTS! I knew I was forgetting something :D

frontend/utils/pages/redirect.ts
```diff
@@ -0,0 +1,16 @@
+import { ParsedUrlQuery } from "querystring";
+import { GetServerSidePropsContext } from "next";
+
+/// Inspired by the api of the `next-redirect` package, but I couldn't find any typings
+/// for that, so I made one myself.
+export const redirect = (
```
frigus02, line 6: ❤️

PR #104, opened 2020-08-11
Event model generation using JSON Schema

frigus02:
Closes [AB#2137](https://dev.azure.com/futurenhs/75407963-f812-43e5-a734-1059bbc036a3/_workitems/edit/2137)

This adds a JSON Schema for event models and TypeScript and Rust type generation.


event-models/rust/src/lib.rs
```diff
@@ -0,0 +1,200 @@
+use chrono::{DateTime, Utc};
```
alsuren, line 1: how much of this file is autogenerated?

event-models/rust/src/main.rs
```diff
@@ -0,0 +1,7 @@
+fn main() -> Result<(), Box<dyn std::error::Error>> {
+    let event = fnhs_event_models::parse()?;
+    println!("Event: {:?}", event);
+    let event = fnhs_event_models::custom_parse()?;
+    println!("Custom Event: {:?}", event);
```
alsuren, line 5: I feel like these should be unit tests, and run using `cargo test`.

event-models/rust/src/lib.rs
frigus02: It's not. It's manually written, which makes this annoying. The `schemafy` crate has a macro, which generates structs from a JSON Schema. But it doesn't support enums with 2 tags (eventType and dataVersion).

I'm interested how you feel about this. We may be able to write a custom macro to generate the file 🤔

event-models/rust/src/main.rs
frigus02: Absolutely agree. If this general approach works for us, we need to clean this up and write some proper tests.

event-models/rust/src/lib.rs
alsuren: crazy idea: dataversion is a string. Could we just include the event type + dataversion in the dataversion field, and switch on that?

event-models/schema.json
```diff
@@ -0,0 +1,101 @@
...
+    "LoggedInEventV2": {
+      "type": "object",
+      "properties": {
+        "eventType": {
+          "type": "string",
+          "enum": ["loggedin"]
+        },
+        "dataVersion": {
+          "type": "string",
+          "enum": ["2"]
```
alsuren, line 62:
```suggestion
          "enum": ["LoggedInEventV2"]
```

event-models/rust/src/lib.rs
frigus02: I was wondering about that as well. We could ignore one of the fields.

In this case the file could look like this: https://github.com/FutureNHS/futurenhs-platform/blob/21556aba477b621c75ed19a2e61d7cdf5f6d8cb3/event-models/rust/src/lib.rs

Serde can come close with adjacently tagged enums + flatten

event-models/schema.json
frigus02: We probably want to ignore the `dataVersion` and put `LoggedInEventV2` in the `eventType`. Because we're able to filter on that.

alsuren: I think I would prefer to filter on eventType == LoggedInEvent and then parse based on dataVersion. That way, if we forget to upgrade some service we will get explosions in our logs rather than unnerving silence.

frigus02: Oooooh. I did not think of that

event-models/rust/src/lib.rs
frigus02: Well okay. Both fields are definitely mandatory. So we need at least a way of serializing them both.

event-models/schema.json
frigus02: Sounds like a good idea. And we can add some simple linter on the schema, that ensures that `dataVersion.startsWith(eventType)` or something similar.

event-models/rust/src/custom.rs
```diff
@@ -0,0 +1,203 @@
...
+                                    serde_json::from_value(data).map_err(de::Error::custom)?,
+                                ),
+                            )*
+                            (a, b) => {
+                                return Err(de::Error::unknown_variant(
+                                    &format!("eventType: {}, dataVersion: {}", a, b),
+                                    DATA_VARIANTS,
+                                ))
+                            }
+                        };
```
alsuren, line 130: The macro is mostly scary because it is huge.

it feels like this pair of match blocks is the only thing in the whole `impl Deserlialize` that actually uses the macro args. If we go down this route, I would be in favour of generating `from_wire(event_type, data_version) -> ...` and `to_wire(...) -> (event_type, data_version)` inside the macro and then writing `impl<'de> Deserialize<'de> for Event` and `impl Serialize for Event` in terms of them, outside the macro.

I'm assuming that this is mostly copy-pasta from some generated code (or cargo-expand output) with modifications. Might be worth calling that out in a comment, and say which bit is actually scary.

event-models/rust/src/using_data_version.rs
```diff
@@ -0,0 +1,94 @@
...
+use serde::{Deserialize, Serialize};
+
+mod gen {
+    use serde::{Deserialize, Serialize};
+
+    schemafy::schemafy!("../schema.json");
+}
+
+// TODO: This would have to be generated. Possibly with some kind of macro
+pub use gen::{LoggedInEventData, LoggedInEventV2Data, LoginFailedEventData};
```
alsuren, line 12: generate the exports in the custom.rs macro, and then pub use `gen::*;` here? (maybe add an extra scope so that you're only importing these generated types?)

alsuren: oops. Ignore me. Didn't read your direct message properly

frigus02: Ah sorry, I did not make that very clear. Yeah, this file is the version that I don't like very much.

event-models/rust/src/custom.rs
frigus02: Yeah, we should be able to refactor this to make the actual macro smaller and easier to read. I'm going to give this a go tomorrow and see how readable we can make this.

frigus02: I refactored and made it nicer.

Offline you also mentioned code generation without macros, i.e. string building/templating + rustfmt. That could be nice but the JSON Schema is quite complex, I find. I'm not sure what the easiest option would be. The easiest thing might be to only generate code for the macro invocation:

```rust
event_serialization!(
    ("loggedin", "1") => LoggedIn(LoggedInEventData),
    ("loggedin", "2") => LoggedInV2(LoggedInEventV2Data),
    ("loginfailed", "1") => LoginFailed(LoginFailedEventData),
);
```

It feels like we could do that later on, though.

event-models/rust/src/client.rs
```diff
@@ -0,0 +1,63 @@
+use crate::Event;
+
+#[derive(Debug, Clone)]
+pub struct Client {
+    url: String,
+    key: String,
+}
```
alsuren, line 7: can you move `struct Client {` closer to `impl Client {`? (either move it down the page or move PublishEventsError and its impls into another module?)

event-models/rust/src/client.rs
```diff
@@ -0,0 +1,63 @@
...
+    /// See also: https://docs.microsoft.com/en-us/rest/api/eventgrid/dataplane/publishevents/publishevents
+    pub async fn publish_events(&self, events: &[Event]) -> Result<(), PublishEventsError> {
+        let res = surf::post(&self.url)
+            .set_header("aeg-sas-key", &self.key)
+            .body_json(&events)?
+            .await?;
+        if res.status() == 200 {
+            Ok(())
+        } else {
+            Err(PublishEventsError::Server(res.status().as_u16()))
```
alsuren: do we trace something about the error before we drop it on the floor?

frigus02: That's probably a good idea. I feel like we should use `opentelemetry` here and properly trace the HTTP request.

event-models/rust/src/models.rs
```diff
@@ -0,0 +1,194 @@
...
+                        }
+                    )*
+                })
+            }
+        }
+    }
+}
+
+event_serialization!(
+    ("ContentView", "1") => ContentView(ContentViewEventData),
```
alsuren, line 141: I like your style with the `=>` :D

stretch goal: I wonder if it would be possible to assert that this list of enum variants matches the list of variants from schemafy::schemafy!("../schema.json"); Is there some kind of exported enum from schemafy that we could match on and check for exhaustiveness or something? (it may already do this. I've not tried).

hello-world/src/main.rs
```diff
@@ -5,14 +5,22 @@ use tracing_subscriber::Registry;
...
-    let batch_exporter =
-        BatchSpanProcessor::builder(exporter, tokio::spawn, tokio::time::interval).build();
-    let provider = sdk::Provider::builder()
-        .with_batch_exporter(batch_exporter)
-        .build();
+    let provider = match env::var("INSTRUMENTATION_KEY") {
+        Ok(instrumentation_key) => {
+            let exporter = opentelemetry_application_insights::Exporter::new(instrumentation_key);
+            let batch_exporter =
+                BatchSpanProcessor::builder(exporter, tokio::spawn, tokio::time::interval).build();
```
alsuren, line 12: this feels unrelated.

alsuren (approved):

event-models/rust/src/models.rs
frigus02: I agree. It would be really nice if we could validate this. Unfortunately `schemafy` does not seem to support enums in general.

I feel like we should either extend schemafy or maybe make a code generator out of string concatenation + rustfmt.

hello-world/src/main.rs
frigus02: Yep, it is. I wanted to test the client with hello world and it annoyed me that I needed an instrumentation key. Happy to move this to a different PR, though.

alsuren (approved):

hello-world/src/main.rs
frigus02: See #190

PR #105, opened 2020-08-12
create terraform  file to build analytics components

cobwebfx:
## Acceptance Criteria

Acceptance criteria:
- [x] Events emitted by other parts of the application are sent to central analytics storage

Tracking requirements:
- [ ] None

Desired outcomes:
- [x] Running in production


infrastructure/environments/production/variables.tf
```diff
@@ -3,11 +3,16 @@ variable "location" {
...
-variable "ip_whitelist_insights" {
+variable "ip_whitelist_analytics" {
   description = "List of whitelisted IPs for use with Synapse"
   type        = map(string)
 }
 
+variable "enable_analytics" {
+  description = "Should analytics be enabled? true or false"
+  type        = bool
+}
```
frigus02, line 14: I don't think we're using this variable in production.

infrastructure/README.md
```diff
@@ -107,7 +107,7 @@ as opposed to sharing a staging environment.
 1. After verifying the plan above, apply changes. The infrastructure will be created in Azure.
```
alsuren, line 107: ^ The above `terraform plan` will fail until `module.platform` is applied. I don't think that anyone runs that step when setting up their cluster. Let's delete it.

(this is my fault anyway, for not updating the docs when I updated the code. Sorry)

infrastructure/modules/platform/analytics.tf
```diff
@@ -0,0 +1,215 @@
...
+  shared_access_policy_name    = "RootManageSharedAccessKey"
+
+  serialization {
+    type     = "Json"
+    encoding = "UTF8"
+  }
+}
+
+resource "azurerm_storage_account" "analytics" {
+  name                     = "analytics${random_string.random_string_10_chars.result}"
```
frigus02, line 69: I don't think we need random characters for the storage account name. Deleting is instant as far as I know.

This means we could use the environment name to make the name unique and more readable, e.g.:

```
fnhsanalytics${var.environment}
```

infrastructure/modules/platform/main.tf
```diff
@@ -130,6 +130,12 @@ resource "random_password" "postgresql_password" {
   upper   = true
 }
 
+resource "random_string" "random_string_10_chars" {
```
alsuren, line 133: seems unsused

infrastructure/modules/platform/analytics.tf
```diff
@@ -0,0 +1,215 @@
...
+  storage_container_name    = azurerm_storage_container.analytics[0].name
+  path_pattern              = "analytics"
+  date_format               = "yyyy-MM-dd"
+  time_format               = "HH"
+
+  serialization {
+    type            = "Csv"
+    encoding        = "UTF8"
+    field_delimiter = ","
+  }
```
frigus02, line 100: In my test I used JSON here. Does CSV work in the same way in Data Factory?

image

infrastructure/modules/platform/analytics.tf
```diff
@@ -0,0 +1,215 @@
...
+  location            = var.location
+  resource_group_name = azurerm_resource_group.platform.name
+  tenant_id           = data.azurerm_client_config.current.tenant_id
+  sku_name            = "standard"
+
+  access_policy {
+    tenant_id = data.azurerm_client_config.current.tenant_id
+    // This object_id relates to the FutureNHS Developers Group ID.
+    // See https://portal.azure.com/#blade/Microsoft_AAD_IAM/GroupDetailsMenuBlade/Overview/groupId/b06ebd00-f52c-4e82-ac88-0520f4320fee
+    object_id = "b06ebd00-f52c-4e82-ac88-0520f4320fee"
```
frigus02, line 130: We have this as a variable here. Let's change that to `var.ad_object_id`

infrastructure/modules/platform/analytics.tf
```diff
@@ -0,0 +1,215 @@
...
+resource "azurerm_sql_firewall_rule" "allow_ip" {
+  for_each            = var.enable_analytics ? var.ip_whitelist_analytics : {}
+  name                = "allow-ip-${each.key}"
+  resource_group_name = azurerm_resource_group.platform.name
+  server_name         = azurerm_mssql_server.synapse[0].name
+  start_ip_address    = each.value
+  end_ip_address      = each.value
+  depends_on = [
+    azurerm_mssql_server.synapse
+  ]
```
frigus02, line 200: Is this "depends_on" necessary? I would have assumed that the dependency is implictely here already, because the `server_name` references the SQL Server resource already.

infrastructure/modules/platform/analytics.tf
```diff
@@ -0,0 +1,215 @@
...
+resource "random_string" "random_string_9_chars" {
+  length  = 9
+  special = false
+}
+
+data "azurerm_client_config" "current" {
+}
+
+resource "azurerm_key_vault" "secret_vault" {
+  name                = "fnhs-analytics-${random_string.random_string_9_chars.result}"
```
alsuren, line 119: I assume this can't be `"fnhs-analytics-${var.environment}"` because there is a grace period between destroying and recreating a key vault with the same name or something? :(

I would be tempted to call it `random_string.key_vault_suffix` or something.

alsuren: ... actually it looks like this is just code that has been moved, and we probably had this discussion already, when it was merged the first time. Happy to leave it as it is.

infrastructure/modules/platform/analytics.tf
```diff
@@ -0,0 +1,215 @@
...
+resource "azurerm_sql_database" "synapse_data_warehouse" {
+  name                = "${var.environment}-synapse-dw"
+  count               = var.enable_analytics ? 1 : 0
+  resource_group_name = azurerm_resource_group.platform.name
+  location            = var.location
+  server_name         = azurerm_mssql_server.synapse[0].name
+
+  //  To upgrade to Synapse, uncomment the below 2 comments
+  //  edition = "DataWarehouse"
+  //  requested_service_objective_name = "DW100c"
```
alsuren, line 184: This means that we have a thing called synapse that isn't actually a synapse db because it's too expensive to justify for dev clusters with no data. Is there a better naming scheme, or do we just live with the lie for now?

 (I suspect the original naming might have been my idea 😊 )

infrastructure/modules/platform/analytics.tf
cobwebfx: Azure is quite strict about only having lowercase letters and numbers for storage. There appears to be a substring command in terraform.  https://www.terraform.io/docs/configuration/functions/substr.html
Maybe we do something like fnhsanalytics${substr(var.environment, 4, -0)}.
That way we remove "dev".  But that assumes that 4 chars is the maximum we'll ever need.

frigus02: Ah, because of the "-" 😞.

Is length really a problem? We have 24 characters. We're already using "fnhstfstate${environment}" for the state bucket storage account. "fnhsanalytics" has 2 more characters. So length-wise I think we're okay.

Should we do `replace(var.environment, "-", "")` to remove the "-"?

cobwebfx: replace(), nice 👍

infrastructure/README.md
```diff
@@ -101,13 +101,13 @@ as opposed to sharing a staging environment.
 1. Create an execution plan:
 
    ```bash
-   terraform plan
+   terraform plan -target module.platform && terraform plan
```
frigus02, line 104: This won't help. The first targeted plan doesn't apply anything, so the second untargeted plan should fail. I think I agree with David. We can probably delete the entire step about creating an execution plan.

infrastructure/modules/platform/analytics.tf
```diff
@@ -0,0 +1,210 @@
...
+  storage_account_key = azurerm_storage_account.analytics[0].primary_access_key
+  storage_container_name = azurerm_storage_container.analytics[0].name
+  path_pattern = "analytics"
+  date_format = "yyyy-MM-dd"
+  time_format = "HH"
+
+  serialization {
+    type = "Json"
+    encoding = "UTF8"
+    format = "Array"
```
frigus02, line 99: I think I tested this with `LineSeparated` instead of `Array`, which worked with data factory. Does `Array` work as well?

infrastructure/README.md
alsuren: `terraform plan -target module.platform` doesn't make any changes to the infrastructure, so  `terraform plan -target module.platform && terraform plan` will still explode on a fresh setup.

Let's really just delete it and rely on the interactive output of apply to show the plan.

infrastructure/modules/platform/analytics.tf
cobwebfx: I only tested blob -> factory -> blob, due to the pipeline not being in the ticket.  You're right. Array is not compatible. I will update it.

infrastructure/modules/platform/analytics.tf
frigus02: Yeah, you're right. Unfortulately the key vault has a grace period when after deletion. So we need to add some kind of randomness to the name.

PR #107, opened 2020-08-13
Send emails fixes

alsuren:
These are the last few items that feel like they're actually in scope for ["Ability to send emails"](https://airtable.com/tblgtzV3sSwquTkIa/viw8ZHoNqvSvl62lt/recVU3BGN3UCVbe1w?blocks=hide). The rest have been punted to ["Register users (service team)"](https://airtable.com/tblgtzV3sSwquTkIa/viw8ZHoNqvSvl62lt/recHKHxau7aAUsB8u?blocks=hide).

## Acceptance Criteria
* `infrastructure/scripts/new-kratos-user.sh dev-david david.laban+dev@red-badger.com` works.
* `infrastructure/scripts/new-kratos-user.sh production david.laban+prod@red-badger.com` :
  * gives the right password-reset link
  * still works if the user has already been created
* Docker image for kratos is built from an actual commit from their master branch, rather than a crazy merge that I did myself.

### DIFF
--- a/infrastructure/kubernetes/kratos/dev-template/kratos.yaml
+++ b/infrastructure/kubernetes/kratos/dev-template/kratos.yaml
@@ -11,3 +11,5 @@ spec:
           env:
             - name: COURIER_SMTP_FROM_ADDRESS
               value: "no-reply@login.fnhs-dev-$NAME.westeurope.cloudapp.azure.com"
+            - name: LOG_LEAK_SENSITIVE_VALUES
+              value: true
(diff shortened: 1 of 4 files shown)


infrastructure/kubernetes/kratos/dev-template/kratos.yaml
```diff
@@ -11,3 +11,9 @@ spec:
           env:
             - name: COURIER_SMTP_FROM_ADDRESS
               value: "no-reply@login.fnhs-dev-$NAME.westeurope.cloudapp.azure.com"
+            - name: LOG_LEAK_SENSITIVE_VALUES
+              value: "true"
```
frigus02, line 15: I assume that was only for debugging purposes.

alsuren: Yes. Any objections to me keeping it (dev-only) or should I revert it and add it only when needed?

frigus02: Hm, 🤔 

Yeah, let's keep it. Might be helpful as long as we're still figuring out some of the quirks of Kratos.

PR #108, opened 2020-08-14
cd test && yarn upgrade

alsuren:
Dependabot is not brave enough to upgrade mocha, so it refuses
to upgrade serialize-javascript.

See https://github.com/FutureNHS/futurenhs-platform/network/alert/test/yarn.lock/serialize-javascript/open
for details.

If everything still works after this PR then let's go with that.
If anything breaks then we'll have to find a gentler way to
upgrade serialize-javascript.

### DIFF
--- a/test/yarn.lock
+++ b/test/yarn.lock
@@ -2,7 +2,7 @@
 # yarn lockfile v1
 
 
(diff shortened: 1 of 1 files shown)


frigus02:
I ran both Cypress and Browserstack tests against production. Both worked perfectly.

PR #109, opened 2020-08-14
use real certificate authority for dev environments

alsuren:


alsuren:
Closing to keep myself focussed.

I couldn't get this to work on my cluster and I don't know enough about how ingress and cert-manager work to debug it in a reasonable amount of time. If this annoys then have a stab at it, but definitely do so in a timebox.

PR #110, opened 2020-08-14
Automatic updated using Dependabot

frigus02:
What do you all think about having automated dependency updates using Dependabot?

- It supports everything we need at the moment: https://docs.github.com/en/github/administering-a-repository/configuration-options-for-dependency-updates#package-ecosystem
- It creates one PR per dependency update, which can be quite noisy. We should update everyone once manually before we merge, otherwise we're going to get flooded.
- There is no way to automerge, unless we have a separate merge bot. I also tested Renovate separately. But I found the configuration a bit harder and the integration with GitHub not as good.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,45 @@
+# https://docs.github.com/en/github/administering-a-repository/keeping-your-dependencies-updated-automatically
+version: 2
+updates:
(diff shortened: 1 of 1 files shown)


frigus02:
We should probably merge #108 first. And do another pass on the frontend and rust dependencies.

PR #112, opened 2020-08-14
Update hello world dependencies

frigus02:
Dependency of #110


frigus02:
Tested on my dev-cluster

PR #113, opened 2020-08-14
Update workspace service dependencies

frigus02:
Dependency of #110


frigus02:
Also tested on dev cluster

PR #132, opened 2020-08-19
initial FusionAuth + NextAuth.js integration

alsuren:
TODO:

~FusionAuth~
- [ ] ~See if it's possible to make the "application" in FusionAuth and extract its oauth secrets automatically (or inject them into FusionAuth using a sealed-secret?)~
  - [ ] ~If this is not possible with FusionAuth then this might be enough to swing us over the edge into deploying a global azure active directory b2c service, and hardcoding app configs as sealed secrets.~
- [ ] ~work out how to style the forms in FusionAuth~
- [ ] ~work out how to invite users in FusionAuth~
- [ ] ~set everything up so that it isn't hard-coded to dev-david~
- [ ] ~punch holes in the ingress controller, so that everything is served from the same domain (remember to rotate the oauth secret before this point)~
- [ ] ~enable 2 replicas again~
- [ ] ~disable self-registration again~
- [ ] ~Run migration scripts in an initcontainer, as a workaround, because of https://github.com/FusionAuth/fusionauth-issues/issues/95~

Azure Active Directory B2C

- [x] Can you make a single app, that serves all of the dev-clusters?
- [x] Can you promote login flows from dev to prod?
- [ ] [OTHER TICKER] Make the password reset link from login page work.
- [x] Can we serve the login flow from the same domain as the website ([login.]beta.future.nhs.uk)?
This is currently impossible. See:
https://docs.microsoft.com/en-us/azure/active-directory-b2c/faq?tabs=app-reg-ga
However, it is on their roadmap, to be completed at the end of 2020/start of 2021. See:
https://feedback.azure.com/forums/169401-azure-active-directory/suggestions/15334317-customer-owned-domains
- [x] Are multi-factor authentication options enough?
- [ ] [OTHER TICKER] Customize login template
- [x] send email via sendgrid (possible)
- [x] Give everyone of the team access

Frontend login sessions
- [ ] Setup DB for session cookies
- [ ] Passport: Integrate passport into Next.js app --> https://github.com/vercel/next.js/blob/canary/examples/custom-server-express/server.js
- [ ] NextAuth.js: make AAD B2C the default provider in NextAuth.js, and don't force the user to click a button?
- [ ] NextAuth.js: Skip "do you want to sign out" button
- [ ] NextAuth.js: look into how to extend the lifetime of session cookies?
  - extra funtimes on the NextAuth.js session refreshing:
    - calling getSession() on the frontend will fetch /session which refreshes the cookie for you.
    - Calling getSession() on the backend also fetches /session (yes: calling over http to itself I think, or even potentially via the load balancer), which will set the cookie, but then there is nothing that then propagates the refreshed cookie from that fetch response up to the frontend.

Kratos
- [ ] Remove Kratos

### DIFF
--- a/aad-b2c-policies/DisplayControl_PasswordReset.xml
+++ b/aad-b2c-policies/DisplayControl_PasswordReset.xml
@@ -1,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
+<TrustFrameworkPolicy
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xmlns:xsd="http://www.w3.org/2001/XMLSchema"
+  xmlns="http://schemas.microsoft.com/online/cpim/schemas/2013/06"
+  PolicySchemaVersion="0.3.0.0"
+  TenantId="futurenhsplatform.onmicrosoft.com"
+  PolicyId="B2C_1A_DisplayControl_PasswordReset"
+  PublicPolicyUri="http://futurenhsplatform.onmicrosoft.com/B2C_1A_DisplayControl_PasswordReset">
+
+  <BasePolicy>
+    <TenantId>futurenhsplatform.onmicrosoft.com</TenantId>
+    <PolicyId>B2C_1A_DisplayControl_TrustFrameworkExtensions</PolicyId>
+  </BasePolicy>
+
+  <RelyingParty>
+    <DefaultUserJourney ReferenceId="PasswordReset" />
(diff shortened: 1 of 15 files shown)


frigus02:
FusionAuth
- Nicer management UI
- Very flexible UI customization
- Optionally hosted: ($425 + $125) / month

AAD B2C
- Fully managed and hosted
- Requires JavaScript on login page
- Clunky management UI
- Basic UI customization (probably enough for us)
- Login happens on futurenhsplatform.b2clogin.com

frigus02:
Succeeded by #133

PR #141, opened 2020-08-24
Create workspace form

matt-paul:
[Ticket](https://airtable.com/tblgtzV3sSwquTkIa/viw8ZHoNqvSvl62lt/recPIOuTjtlaFKtyn?blocks=hide) [AB#1764](https://dev.azure.com/futurenhs/75407963-f812-43e5-a734-1059bbc036a3/_workitems/edit/1764)

[Zeplin]()

## Acceptance Criteria

(from ticket: Development or Service team are able to set up workspaces when requested to do so or when we need to set up workspaces to import
)

## Notes
- An [Apollo Federation Gateway](https://www.apollographql.com/docs/apollo-server/federation/gateway/) has been setup to pass through graphql requests from the frontend.  The Graphql playground can be found at `http://localhost:4455/api/graphql` which can be used to query the workspace service (and any future services).

- At present, we are using [Graphql-request](https://github.com/prisma-labs/graphql-request) as the client on the frontend, there may well be a time whenn we need to change this to something more fully featured (caching etc) such as Apollo Client (and that my be before this PR is finished, lets see...).

For code generation, set up port forwarding for the workspace service:
`kubectl port-forward -n workspace-service <MY_LOCAL_WORKSPACE_POD_NAME> 8080:3030`

Add any new queries to .graphql files

## Pre-review checklist:

- [ ] Tests

- [ ] Documentation

- [ ] Feature flag

## TODO

Rust:

- [x] add description field on workspace
- [x] enable federation on schema (
- [x] should `fn workspace(..., id)` be annotated with `#[field(description = "...")]`
- [ ] Make database schema use VARCHAR($LIMIT) to match business requirements? (or is input validation enough? :/ )

Infra:

- [ ] make workspace service url configurable in frontend using environment variable (default to http://workspace-service.workspace-service/ ?)
- [ ] make kubernetes service for workspace service if it doesn't already exist.

Frontend:

- [x] regenerate the new schema (`yarn generate` while the server is running) (maybe dig into flakeyness if it occurs again)
- [x] make the form work
- [x] implement submit button
- [ ] how do we do session management in apollo-server-micro? (Jan pointed out https://graphql.org/learn/authorization/ )

## Testing information

- Test plan?

## Test:

- [ ] Tested locally

- Platforms/Browsers:

  - [ ] Internet Explorer 11

  - [ ] Edge

- Accessibility:

  - [ ] Screen Reader

  - [ ] Keyboard Navigation

  - [ ] Magnification

  - [ ] Contrast

  - [ ] Text size 200%

  - [ ] Landscape (Mobile)


frontend/package.json
```diff
@@ -10,20 +10,30 @@
...
   "dependencies": {
+    "@apollo/federation": "^0.19.1",
+    "@apollo/gateway": "^0.19.1",
     "@azure/monitor-opentelemetry-exporter": "^1.0.0-preview.3",
     "@microsoft/applicationinsights-web": "^2.5.8",
     "@opentelemetry/node": "^0.10.2",
     "@opentelemetry/plugin-https": "^0.10.2",
     "@opentelemetry/tracing": "^0.10.2",
     "axios": "^0.20.0",
+    "@oryd/kratos-client": "^0.4.6-alpha.1",
```
matt-paul, line 27: hmm this seems like it's slipped through in the rebase, i'm not sure we use kratos now?

frontend/components/Header/Header.tsx
```diff
@@ -30,19 +24,18 @@ const StyledImg = styled.img`
   `}
 `;
 
-const Header = ({
-  imageRight,
-  imageRightURL,
-  imageRightAltText,
-}: HeaderProps) => {
+const Header = () => {
```
matt-paul, line 27: Need to double check with Kim as there are designs that have a different image so this change may not work if the designs are current.

frontend/pages/admin/workspaces.tsx
```diff
@@ -0,0 +1,138 @@
...
+  ${({ theme }) => `
+  border-top: 1px solid ${theme.colorNhsukGrey1};
+  padding-top: 24px;
+  margin-bottom: 8px;
+  `}
+`;
+
+const FormField = styled.div`
+  padding-bottom: 40px;
+  #text {
```
matt-paul, line 45: Fixme

frontend/components/Header/Header.tsx
matt-paul: This is correct, we just have a single one. The confusion came because there are designs showing the new image which will replace the old one across the site

.github/workflows/linting.yaml
```diff
@@ -50,7 +50,7 @@ jobs:
       - name: Install modules
         run: yarn
       - name: Run Prettier
-        run: yarn prettier --check "**/*.{js,ts,tsx,md,yaml,graphql}"
+        run: yarn prettier --check "**/*.{js,ts,tsx,md,yml,yaml,graphql}"
```
trcywu, line 53: I remember earlier on in the project we decided to use .yaml consistently and not use .yml. Should we stick to it?

frontend/components/Head/Head.tsx
```diff
@@ -0,0 +1,13 @@
...
+interface Props {
+  title: string;
+}
+const PageHead = ({ title }: Props) => (
+  <Head>
+    <title>{title}</title>
+  </Head>
+);
+
+export default PageHead;
```
trcywu, line 13: is this component 'PageHead' or 'Head'?

.github/workflows/linting.yaml
matt-paul: Yes. This is here just to allow prettier to verify that the all yaml (whether yaml or yml ) are correct if they end up in the repo for whatever(accidental) reason. I think it's helpful for prettier to verify everything. As time goes on i imagine other things will slip through, there is dependabot file in there already.  As a separate point though, the codegen.yml was originally copied from the example which was yml, i will change it thanks!

frontend/components/Head/Head.tsx
matt-paul: We are importing the default import <Head/> from next, so can't use it in the file.  However, in terms of the rest of the code base, i think it's more helpful and resulting in clearer pages if we simplify component names where possible, therefore to the outside world the component is <Head /> as no one needs to really care of the origninal.

matt-paul: (if it wasn't default then i would have done `import { Head as NextHead....} or something

Cargo.toml
```diff
@@ -0,0 +1,6 @@
+[workspace]
```
frigus02, line 1: I wonder if having a Cargo workspace is a good idea.

- How does this play with the Cargo.lock files we have in the hello-world and workspace-service folders. Should we remove them?
- Does this make CI more difficult? Because of the lock file we're sharing exact version constraints between services. Does that means when the global Cargo.lock file changes we have to run the CI build for hello-world and workspace-service?

@alsuren What are your thoughts about this?

frontend/graphql.schema.json
```diff
@@ -0,0 +1,1408 @@
+{
```
frigus02, line 1: Is this file auto-generated? I'm not sure if we should commit this. Maybe we should add this to .gitignore.

frontend/lib/generated/graphql.ts
```diff
@@ -0,0 +1,222 @@
+import { GraphQLClient } from "graphql-request";
```
frigus02, line 1: This is definitely generated. We should probably add the entire folder to .gitignore.

frontend/package.json
```diff
@@ -10,20 +10,30 @@
...
     "@microsoft/applicationinsights-web": "^2.5.8",
     "@opentelemetry/node": "^0.10.2",
     "@opentelemetry/plugin-https": "^0.10.2",
     "@opentelemetry/tracing": "^0.10.2",
+    "apollo-server-micro": "^2.16.1",
     "axios": "^0.20.0",
     "azure-eventgrid": "^1.6.0",
+    "classnames": "^2.2.6",
     "connect-pg-simple": "^6.2.1",
+    "cookie": "^0.4.1",
```
frigus02, line 31: I don't think the cookie package is needed.

workspace-service/src/lib.rs
```diff
@@ -29,9 +32,11 @@ pub fn log<'a>(
 }
 
 pub async fn create_app(connection_pool: PgPool) -> anyhow::Result<Server<graphql::State>> {
+    let cors = CorsMiddleware::new().allow_origin(Origin::from("*"));
     let mut app = tide::with_state(graphql::State::new(connection_pool));
 
     app.with(log);
+    app.with(cors);
```
frigus02, line 39: I don't think we need a CORS middleware here, since the service is not called directly from the browser.

workspace-service/migrations/20200901103018_add-long_description.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE workspace ADD COLUMN long_description TEXT NOT NULL DEFAULT '';
```
frigus02, line 1: Not super important, but should we just call the column `description`?

frontend/pages/admin/workspaces.tsx
```diff
@@ -0,0 +1,154 @@
...
+}
+
+const PageContent = styled.div`
+  ${({ theme }) => `
+  background-color: ${theme.colorNhsukWhite};
+  min-height: 100vh;
+  padding-top: 24px;
+  padding-left: 10%;
+  padding-right: 10%;
+  h3 {
```
frigus02, line 40: Is the h3 style still needed?

frontend/pages/admin/workspaces.graphql
```diff
@@ -0,0 +1,33 @@
...
+
+query WorkspaceByIDQuery($id: ID!) {
+  workspace(id: $id) {
+    id
+    title
+    longDescription
+  }
+}
+
+mutation CreateWorkspaceMutation($title: String!, $longDescription: String!) {
```
frigus02, line 17: I think this is the only muration/query we're using from this file. I'd say we remove the other ones and only have the things in here, that we're actually using.

Cargo.toml
matt-paul: @alsuren is behind that one !

workspace-service/migrations/20200901103018_add-long_description.sql
matt-paul: description seemed to highlight in a weird way in vs code when i was working on this way with David, in a way that made it seem like it was a reserved word..happy to change it if its fine though, it was annoying to call it that.  I wonder why vscode didn't like it though..

frontend/pages/admin/workspaces.tsx
matt-paul: goog point, we changed the levels of the headings

workspace-service/migrations/20200901103018_add-long_description.sql
frigus02: Fair enough. I tought `long_description` is longer to type without any major advantage. But if just `description` requires special escaping, probably fine to leave it like this

workspace-service/src/lib.rs
matt-paul: Ah that was from local deving earlier in the ticket, good spot

workspace-service/migrations/20200901103018_add-long_description.sql
alsuren: it was just the highlighting wasn't it? Maybe put the queries into tableplus and see how it prettifies them? In the worst case, escaping it like `ALTER TABLE workspace ADD COLUMN "description" TEXT NOT NULL DEFAULT '';` might be fine. I'm not that attached to either name.

Cargo.toml
alsuren: I will take a look on my own machine. My hope was that the lockfiles in the lower directories took priority, but I don't actually know. If the child lockfiles are just ignored then this might be a non-starter.

workspace-service/src/graphql.rs
```diff
@@ -13,92 +13,108 @@ pub struct Workspace {
...
 
-pub struct QueryRoot;
+pub struct Query;
 
 #[Object]
-impl QueryRoot {
-    #[field(description = "Get all Workspaces")]
+impl Query {
+    #[field(desc = "Get all Workspaces")]
     async fn workspaces(&self, context: &Context<'_>) -> FieldResult<Vec<Workspace>> {
```
alsuren, line 62: There is no pagination on this query, which is a bit scary. I'm happy to leave this as it is for now though. I have noted it down on AB2121 as a thing to think about later.

Cargo.toml
frigus02: Yeah, I'm not 100% how workspaces work with Cargo. I assume it improves editor support because we now have a root-level Cargo.toml. And it does help you to align on crate versions, but in our case we might not want that.

workspace-service/src/graphql.rs
matt-paul: Marking as resolved due to ticket having been created.

workspace-service/migrations/20200901103018_add-long_description.sql
matt-paul: As we just discussed, we are leaving it 'as is' , resolving

Cargo.toml
matt-paul: As discussed, removing this for now and we will look for another solution for proper monorepo VSCode support

frontend/components/Head/Head.tsx
trcywu: i think PageHead is more explicit in explaining what it is.

frigus02 (approved):

frontend/components/Head/Head.tsx
matt-paul: I think i would politely disagree with that 🙂.  It's basically a <head> element in a component form, so I think calling it <Head> is explicit as it gets.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/head
https://nextjs.org/docs/api-reference/next/head

workspace-service/src/lib.rs
```diff
@@ -1,7 +1,10 @@
 use sqlx::PgPool;
 use std::future::Future;
 use std::pin::Pin;
-use tide::{Next, Redirect, Request, Response, Server};
+use tide::{
+    security::{CorsMiddleware, Origin},
```
frigus02, line 5: This is still a bit weird. The CorsMiddleware is not used, so I think the change should not be here. But I think it didn't exist locally on your machine when we checked. Can you have another look?

frontend/src/generated/graphql.tsx
```diff
@@ -0,0 +1,56 @@
+export type Maybe<T> = T | null;
```
frigus02, line 1: Is that file needed or is this a duplicate of the lib/generated folder, which is now in .gitignore?

matt-paul: Dammit, rebase fail

frigus02 (approved):

PR #161, opened 2020-09-01
upgrade kubernetes

cobwebfx:
Upgrade Kubernetes to latest version (1.17.9) available in Azure WestEurope that is not in preview.


frigus02 (approved):
Nice! How does this work when you do `terraform apply`? Does it do a cluster upgrade?

cobwebfx:
> Nice! How does this work when you do `terraform apply`? Does it do a cluster upgrade?

Yes, I believe so. The documentation isn't clear, but the dev site doesn't go down during execution.

PR #162, opened 2020-09-01
Gradient container

declanslevin:
[1724 - Gradient Container](https://dev.azure.com/futurenhs/FutureNHS/_workitems/edit/1724) [AB#1724](https://dev.azure.com/futurenhs/75407963-f812-43e5-a734-1059bbc036a3/_workitems/edit/1724)

[Zeplin](https://zpl.io/2ZWoe0E)

## Acceptance Criteria

- [x] There is a gradient bar and right-aligned text on the log in page
- [x] ~~Figure out which grid system is best to use~~ _Used flexbox instead_
- [x] The gradient bar extends the full width of the page

## Pre-review checklist:

- [ ] Tests

- [ ] Documentation

- [ ] Feature flag

## Testing information

- Test plan?

## Test:

- [ ] Tested locally

- Platforms/Browsers:

  - [ ] Internet Explorer 11

  - [ ] Edge

- Accessibility:

  - [ ] Screen Reader

  - [ ] Keyboard Navigation

  - [ ] Magnification

  - [ ] Contrast

  - [ ] Text size 200%

  - [ ] Landscape (Mobile)

### DIFF
--- a/frontend/components/Login/LoginPanel/LoginPanel.tsx
+++ b/frontend/components/Login/LoginPanel/LoginPanel.tsx
@@ -17,27 +17,21 @@ const StyledLoginPanel = styled(Panel)`
     flex-direction: column;
     justify-content: center;
 
-    // .h2 {
-    //   font-size: ${theme.nhsukTypographyScale.tablet};
-    //   line-height: 36px;
-    // }
-
-    // .p {
-    //   font-size: 24px;
-    //   color: #425462;
-    // }
-
     @media (min-width: 1200px) {
       padding-left: 108px;
     }
-    `}
+
+    h2 {
+      color: ${theme.colorNhsukGrey1};
+    }
+  `}
 `;
 
 const LoginPanel = ({ label, text }: LoginPanelProps) => {
   return (
     <StyledLoginPanel>
-      <h2>{label}</h2>
-      <p>{text}</p>
+      <h1>{label}</h1>
+      <h2>{text}</h2>
     </StyledLoginPanel>
   );
 };
--- a/frontend/components/Login/LoginPanel/LoginPanel.tsx
+++ b/frontend/components/Login/LoginPanel/LoginPanel.tsx
@@ -8,13 +8,28 @@ type LoginPanelProps = {
 };
 
 const StyledLoginPanel = styled(Panel)`
-  ${() => `
+  ${({ theme }) => `
     background-color: transparent;
     padding: 0px;
-    max-width: 392px;
+    width: 100%;
+
     display: flex;
     flex-direction: column;
     justify-content: center;
+
+    // .h2 {
+    //   font-size: ${theme.nhsukTypographyScale.tablet};
+    //   line-height: 36px;
+    // }
+
+    // .p {
+    //   font-size: 24px;
+    //   color: #425462;
+    // }
+
+    @media (min-width: 1200px) {
+      padding-left: 108px;
+    }
     `}
 `;
 
--- a/frontend/components/Login/LoginPanel/LoginPanel.tsx
+++ b/frontend/components/Login/LoginPanel/LoginPanel.tsx
@@ -1,0 +1,26 @@
+import React from "react";
+import { Panel } from "nhsuk-react-components";
+import styled from "styled-components";
+
+type LoginPanelProps = {
+  label: string;
+  text: string;
+};
+
+const StyledLoginPanel = styled(Panel)`
+  ${({ theme }) => `
+    background-color: ${theme.colorNhsukWhite};
+    padding: 15px;
+    `}
+`;
+
+const LoginPanel = ({ label, text }: LoginPanelProps) => {
+  return (
+    <StyledLoginPanel>
+      <h2>{label}</h2>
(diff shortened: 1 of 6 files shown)


frontend/pages/gradient.tsx
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @media (min-width: 769px) {
+      min-width: 477px;
+    }
+
+    @media (min-width: 990px) {
+      min-width: 481px;
+      min-height: 504px;
+    }
```
frigus02, line 27: We might want to refactor extracting the breakpoint variables later on to something like this:

```ts
export const mq = (
  breakpoint: "mobile" | "tablet" | "desktop" | "large-desktop"
) => ({ theme }) => `min-width: ${theme.mqBreakpoints[breakpoint]}`;

export const StyledLoginPlaceholder = styled.div`
  ${({ theme }) => `
    @media (${mq("mobile")}) {
      min-width: 435px;
    }
  `}
`;
```

frontend/components/Login/LoginPanel/LoginPanel.tsx
```diff
@@ -0,0 +1,39 @@
...
+    padding: 0px;
+    width: 100%;
+
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+
+    @media (min-width: 1200px) {
+      padding-left: 108px;
+    }
```
frigus02, line 22: At the moment the login panel makes an assumption that it's only used inside a component that makes it side-by-side another component at this breakpoint. That's okay in this case because both the LoginPanel and the Gradient are only going to be used together on the login page. But in general it's better if components don't assume anything about their parent, because it makes them easier to reuse and harder to accidentally break.

I think this is where CSS Grid is actually quite nice. Because you can define the gap in the grid.

Don't need to change anything now. But something to keep in mind for the future.

frontend/pages/gradient.tsx
```diff
@@ -0,0 +1,60 @@
+import React from "react";
+import { PageLayout } from "../components/PageLayout";
+import { Header } from "../components/Header";
+import { Gradient } from "../components/Gradient";
+import { LoginPanel } from "../components/Login/LoginPanel";
+import styled from "styled-components";
+
+export const StyledPageWrapper = styled.div`
+  ${({ theme }) => `
+    height: 100vh;
```
frigus02, line 10: I think that should probably be `min-height`, to make sure it scrolls properly.

frontend/pages/gradient.tsx
```diff
@@ -0,0 +1,60 @@
...
+    width: 100%;
```
frigus02, line 11: I think `width: 100%;` should be the default for divs. Do we need this?

frontend/pages/gradient.tsx
```diff
@@ -0,0 +1,60 @@
...
+    background-color: ${theme.colorNhsukWhite};
+  `}
+`;
+
+export const StyledLoginPlaceholder = styled.div`
+  ${({ theme }) => `
+    height:100%;
```
frigus02, line 18: Do we need `height: 100%`? Will that prevent scrolling?

frigus02 (commented):
Some smaller comments. Looks really good, though.

Can you make this "ready for review"? It's still a Draft PR.

frontend/pages/gradient.tsx
declanslevin: On this point, I've used vh to force it to be full height for now, but this isn't the best solution because if the user has reduced the height of their browser window, when you scroll down, the bottom edge of the div becomes visible and you can see the grey background.

I _think_ the solution is to set html and body to 100% height but I wasn't sure how to go about that in case it screws with the default NHS styling.

frontend/pages/gradient.tsx
declanslevin: I think as long as the parent container is set to 100% then it's fine? (I had to do it for the gradient or it doesn't fill the full width of the page)

frontend/pages/gradient.tsx
declanslevin: Yeh it's not necessary because there's a min-height on the placeholder

frontend/pages/gradient.tsx
frigus02: Yeah, maybe we want to add `min-height: 100vh` to the body instead. I would not add a fixed `height` because it means there is no top level scrolling anymore.

frontend/components/Login/LoginPanel/LoginPanel.tsx
declanslevin: Yeh I was aware that this would be bad practise normally, but as it was for a one-off component that was always only going to appear on the login page I thought it would be ok

frigus02: Yep, it's definitely fine for this one. We could move the `Gradient` component into the `frontend/components/Login` folder to make this relationship clearly visible. But also here, I'm happy to do this later.

trcywu (approved):

frigus02 (approved):

PR #184, opened 2020-09-07
Page focus accessibility

trcywu:
[Ticket]()

[Zeplin]()

## Acceptance Criteria

- Added aria-live label to ensure focus to H1 upon navigating to new page
- Made a MainHeading component which should be used for all h1 headings in page

## Pre-review checklist:

- [ ] Tests

- [ ] Documentation

- [ ] Feature flag

## Testing information

- Test plan?

## Test:

- [ ] Tested locally

- Platforms/Browsers:

  - [ ] Internet Explorer 11

  - [ ] Edge

- Accessibility:

  - [ ] Screen Reader

  - [ ] Keyboard Navigation

  - [ ] Magnification

  - [ ] Contrast

  - [ ] Text size 200%

  - [ ] Landscape (Mobile)

### DIFF
--- a/frontend/components/Login/LoginPanel.tsx
+++ b/frontend/components/Login/LoginPanel.tsx
@@ -1,6 +1,7 @@
 import React from "react";
 import { Panel } from "nhsuk-react-components";
 import styled from "styled-components";
+import { MainHeading } from "../../components/MainHeading";
 
 type LoginPanelProps = {
   label: string;
@@ -29,7 +30,7 @@ const StyledLoginPanel = styled(Panel)`
 const LoginPanel = ({ label, text }: LoginPanelProps) => {
   return (
     <StyledLoginPanel>
-      <h1>{label}</h1>
+      <MainHeading>{label}</MainHeading>
       <h2>{text}</h2>
     </StyledLoginPanel>
   );
--- a/frontend/components/MainHeading/MainHeading.tsx
+++ b/frontend/components/MainHeading/MainHeading.tsx
@@ -1,0 +1,15 @@
+import React from "react";
+interface Props {
+  children: string;
+}
+
+/**
+ * This component should be used on every page as the main heading.
+ * It's a workaround for a known next.js accessibility issue that doesn't read out
+ * headlines on navigation
+ */
+const MainHeading = ({ children }: Props) => {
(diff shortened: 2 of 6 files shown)


frigus02 (approved):

alsuren (approved):

frigus02:
Replaces #171

frontend/components/Login/LoginPanel.tsx
```diff
@@ -29,7 +30,7 @@ const StyledLoginPanel = styled(Panel)`
 const LoginPanel = ({ label, text }: LoginPanelProps) => {
   return (
     <StyledLoginPanel>
-      <h1>{label}</h1>
+      <MainHeading>{label}</MainHeading>
```
matt-paul, line 33: Slight preference for calling this component <H1>, just so that it's clear for dev users what it is underneath.  In this case it's pretty obvious, but if we have components down the line that are essentially h2, h3, h4 etc, then it would save us having to think about 'SubHeading', 'SubsubHeading' or whatever naming terminology we might come up with, and having to 'convert name to html heading element'.  What are you thoughts on that?

frigus02: Interesting. My feeling was that calling it `H1` is possibly misleading. If a page has multiple `h1` elements on a page, it should still only have one `MainHeading`. Because that one will cause the screen reader to read the headline and we only want that once. So while you mostly have only one `h1` element, it's not strictly a 1:1 replacement.

matt-paul: hmm, but a page should never have multiple H1's should it?

matt-paul: (for accessability reasons that you mention)

frigus02: Hm, the document outline matters. In theory you can build that using only h1 elements. E.g.

```html
<main>
  <h1>Top</h1>
  <section>
    <h1>Sub</h1>
  </section>
</main>
```

This exists for a few years now. But browser support was quite varying, if I remember correctly. Not sure how that looks like now.

I'm not actually sure if multiple top level heading are a problem accessibility wise. It feels wrong to me, but I'm not sure how bad it is.

PR #185, opened 2020-09-08
Dynamic assetprefix and instrumentationkey for frontend

frigus02:
## Description

We currently have a problem with our frontend deployment on dev clusters:

- In order to serve a login page template for AAD from our Next.js app, we had to make our asset URLs absolute. We did that using an assetPrefix, which is only set for production builds. Because that's inserted to pages at build time we set it to beta.future.nhs.uk.
- We also deploy production builds to our dev clusters. Those run under a different domain, though.
- This means that assets currently don't load on dev clusters, unless the same asset also exists in production already.

This fixes the problem by:

- Using a placeholder name during build time (`__FNHS_ORIGIN__`)
- Replacing the placeholder with the actual domain on Docker container startup.

Since that adds some form of templating, it also enables us to do the same for the Application Insights instrumentation key. So we do that as well.

### Downside:

This makes it impossible to have a read-only file system in the Docker container. We can probably fix that by doing the placeholder replacement in express.js. But then we take on a runtime cost. We could then shift the load to a CDN, which can cache all the resources. But that's more complex, so let's think about this later.

## Testing information

- Test the site loads on dev clusters and URLs for assets are absolute and use the domain of that dev cluster.
- Also test that it uses the correct instrumentation key.

## Test:

- [x] Tested locally

- [x] Tested on dev cluster

### DIFF
--- a/frontend/Dockerfile
+++ b/frontend/Dockerfile
@@ -12,7 +12,9 @@ COPY . .
 
 # Building app
 RUN yarn test
-RUN NEXT_PUBLIC_INSTRUMENTATION_KEY=18be3b83-4ad5-4186-a7b6-f65dd1253a77 yarn build
+RUN NEXT_PUBLIC_INSTRUMENTATION_KEY=__FNHS_INSTRUMENTATION_KEY__ yarn build
 
 # Running the app
+COPY docker-entrypoint.sh /usr/local/bin/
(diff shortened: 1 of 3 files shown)


alsuren (approved):
looks like what we discussed. I assume you've already tested it.

frigus02:
Yeah, tested locally using `yarn dev` and on my cluster on both static pages and server rendered pages.

PR #187, opened 2020-09-08
AB2134 Store/Retrieve folders in workspaces database

alsuren:
[Ticket]() [AB#2134](https://dev.azure.com/futurenhs/75407963-f812-43e5-a734-1059bbc036a3/_workitems/edit/2134)

Just the db and graphql, plus some smoke tests. Script to populate the cluster db will come later.

As a user
I want to be able to access real folders in the database (rather than dummy data)
so that I can refresh and see any changes that are made

### Notes from 4 Caballeros
* Folders are only required to be a single level deep for now (this may change later?)
* There will be no files at the top-level of a workspace. Only folders.
* Special pages like the homepage/about page/whatever will not be treated like files.
* Turns out that the workspace service has no tests yet. Guess I'll add some for this ticket.

[Zeplin]() None, but it's made to support the following UI:

## Pre-review checklist:

- [x] Tests

- [x] Documentation

- [ ] ~Analytics (user analytics, e.g. Google Analytics)~  no frontend yet

- [ ] ~Observability (metrics/tracing)~ blocked on rust support for events

- [ ] ~Feature flag~

- [x] Data columns adhere to the [Dublin Core Metatdata schema](https://www.dublincore.org/specifications/dublin-core/dcmi-terms/). Further information is available [here](https://www.gov.uk/government/publications/recommended-open-standards-for-government/using-metadata-to-describe-data-shared-within-government) -- This is for exported tabular data. There is no exporting happening here. That said, Dublin Core does describe a `description` field, which it describes as "Description may include but is not limited to: an abstract, a table of contents, a graphical representation, or a free-text account of the resource.". This is enough of an incentive for me to rename long_description to description. ~I will do this as a separate PR though, because I think it will cause conflicts with other work that is ongoing.~

## Testing information

- Test plan?

## Test:

- [x] Tested locally

- Platforms/Browsers:

  - [ ] Internet Explorer 11

  - [ ] Edge


workspace-service/src/graphql.rs
```diff
@@ -14,29 +14,65 @@ pub struct Workspace {
...
+#[SimpleObject(desc = "A folder")]
+pub struct Folder {
+    #[field(desc = "The id of the folder")]
+    id: ID,
+    #[field(desc = "The title of the folder")]
+    title: String,
+    #[field(desc = "The description of the folder")]
+    description: String,
+    #[field(desc = "The workspace that this folder is in")]
+    workspace: ID,
```
frigus02, line 39: I wonder if it's possible to link this to an actual `Workspace` object here. I feel like that would be the GraphQL way somehow. But it's probably not worth looking into this at the moment.

frigus02 (approved):

workspace-service/src/graphql.rs
frigus02: I think ignore me. I tried to find how connections between objects are possible in async-graphql (or in general). And I think structs just have IDs referencing something else, like we have here. And we could them implement a function like this to query the workspace:

```rust
#[async_graphql::Object]
struct Folder {... }

impl Folder {
    async fn workspace(&self, ctx: &Context<'_>) -> Workspace {
        let pool = context.data()?;
        let workspace = db::Workspace::get_by_id(self.workspace, pool).await?;
        Ok(workspace.into())
    }
}
```

And it's definitely not needed at the moment.

workspace-service/Cargo.toml
```diff
@@ -5,19 +5,17 @@ edition = "2018"
...
 tide = "0.13.0"
-tokio = { version = "0.2.22", features = ["macros", "time", "rt-core", "stream"] }
 tracing = { version = "0.1.19", features = ["attributes"] }
 tracing-futures = "0.2.4"
 tracing-opentelemetry = "0.7.0"
 tracing-subscriber = "0.2.11"
 anyhow = "1.0.32"
 async-graphql = "1.17.21"
 async-graphql-tide = "1.17.9"
+async-std = { version = "1.6.3", features = ["attributes", "unstable"] }
```
alsuren, line 19: what features are in "unstable" that we need?

frigus02: async_std::stream::interval, which we use for the opentelemetry exporter. Not great to use unstable features definitely. I'm keeping an eye on the opentelemetry thing to see when something changes.

PR #188, opened 2020-09-09
Enforce order of imports.

matt-paul:
This PR is to create some logical order with our imports - the rules can be found in .eslintrc.json and seem reasonably sane from a react point of view - lifted from what i found other people out there are using.

NOTE: Server.js has a weird order of imports (from eslint perspective), and there doesn't seem massive value in adding extra rules just for that file, so i have ignored it.

What do people think?

### DIFF
--- a/frontend/.eslintrc.json
+++ b/frontend/.eslintrc.json
@@ -21,9 +21,28 @@
     "ecmaVersion": 11,
     "sourceType": "module"
   },
-  "plugins": ["react", "@typescript-eslint"],
+  "plugins": ["react", "@typescript-eslint", "import"],
   "rules": {
     "no-unused-vars": "off",
+    "import/order": [
+      "error",
+      {
+        "groups": ["builtin", "external", "internal"],
+        "pathGroups": [
+          {
+            "pattern": "react",
+            "group": "external",
+            "position": "before"
+          }
+        ],
+        "pathGroupsExcludedImportTypes": ["react"],
+        "newlines-between": "always",
+        "alphabetize": {
+          "order": "asc",
+          "caseInsensitive": true
+        }
+      }
+    ],
     "@typescript-eslint/no-unused-vars": ["error"]
   },
   "settings": {
--- a/frontend/components/Head/Head.tsx
+++ b/frontend/components/Head/Head.tsx
@@ -1,4 +1,5 @@
 import React from "react";
+
 import { default as NextHead } from "next/head";
 
 interface Props {
--- a/frontend/components/Header/Header.tsx
+++ b/frontend/components/Header/Header.tsx
@@ -1,4 +1,5 @@
 import React from "react";
+
 import styled from "styled-components";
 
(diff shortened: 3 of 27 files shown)


frontend/.eslintrc.json
```diff
@@ -21,9 +21,28 @@
...
+      {
+        "groups": ["builtin", "external", "internal"],
+        "pathGroups": [
+          {
+            "pattern": "react",
+            "group": "external",
+            "position": "before"
+          }
+        ],
+        "pathGroupsExcludedImportTypes": ["react"],
```
frigus02, line 38: What does `pathGroups` and `pathGroupsExcludedImportTypes` do?

frigus02 (approved):
Looks good to me

frontend/.eslintrc.json
matt-paul: Puts react at the top

matt-paul: Just because everyone is so used to that convention

frigus02: Ah right. In my mind we don't need that. But it's also fine for me 🙂

alsuren:
If VSCode's format-on-save is capable of applying this then go for it. On our previous project, we would have the problem that  vscode would auto-import something at the bottom of the imports list and then the linter would forbid committing because they weren't in the right order. This was infuriating.

The other thing that was infuriating was that vscode would auto-import a thing when you tab-complete it, and then not auto-remove it when you removed the code. This would also cause a linter error. Maybe I am conflating these two things in my head.

I suppose I mostly want to run eslint --fix on save or something. If that is possible then I'm happy to have as complicated linting rules as you want. Otherwise it will just make me grumpy. That said, I'm not the only person on the team, so if other people are more grumpy about unsorted imports then go for it.

matt-paul:
> If VSCode's format-on-save is capable of applying this then go for it. On our previous project, we would have the problem that vscode would auto-import something at the bottom of the imports list and then the linter would forbid committing because they weren't in the right order. This was infuriating.
> 
> The other thing that was infuriating was that vscode would auto-import a thing when you tab-complete it, and then not auto-remove it when you removed the code. This would also cause a linter error. Maybe I am conflating these two things in my head.
> 
> I suppose I mostly want to run eslint --fix on save or something. If that is possible then I'm happy to have as complicated linting rules as you want. Otherwise it will just make me grumpy. That said, I'm not the only person on the team, so if other people are more grumpy about unsorted imports then go for it.


It nicely orders things on save for >95% of our uses.  The only one it struggled with was server.js.

matt-paul:
```The other thing that was infuriating was that vscode would auto-import a thing when you tab-complete it, and then not auto-remove it when you removed the code. This would also cause a linter error. Maybe I am conflating these two things in my head.```

This is a separate issue.  I have autoimports turned off because it annoyed me and sometimes imported the wrong thing

PR #197, opened 2020-09-10
Selenium e2e tests

alsuren:
Test with `watchexec --watch=test/internet-explorer/ -- ./test/scripts/internet-explorer-bs.sh prod` (or just `./test/scripts/internet-explorer-bs.sh prod` if you don't need a watcher.)

with `./test/scripts/local-bs-connect.sh` running in a background tab.

Includes test for [AB#1766](https://dev.azure.com/futurenhs/75407963-f812-43e5-a734-1059bbc036a3/_workitems/edit/1766) - Workspace homepage 

## Testing information

- Is there any special setup required?

- Test plan?

## Test:

- [x] Tested locally

- Platforms/Browsers:

  - [ ] Google Chrome (latest version)

  - [ ] Internet Explorer 11

  - [ ] Edge

- Accessibility:

  - [ ] Screen Reader

  - [ ] Keyboard Navigation

  - [ ] Magnification

  - [ ] Contrast

  - [ ] Text size 200%

  - [ ] Touch target areas (Mobile)

  - [ ] Landscape (Mobile)

PR #198, opened 2020-09-10
Pass AAD login in Cypress

frigus02:
- Add `login` command to Cypress tests, which performs a login with AAD.
- Load environment variables from .env file in Cypress tests. This contains the username/password used for the AAD login.
- Add test for workspace homepage (see [AB#1766](https://dev.azure.com/futurenhs/75407963-f812-43e5-a734-1059bbc036a3/_workitems/edit/1766))

## Test

- [x] Tested locally

- Platforms/Browsers:

  - [ ] Edge


frigus02:
We should wait for #197 before we merge this one.

alsuren (approved):
hrm. This feels really fragile, but I think that all solutions are going to end up really fragile, and at least we have the selenium tests doing things properly, so we can know when login is *actually* broken.

alsuren:
I say we wait for the current implementation to break and then think about a side-channel login if we need to fix it more than once in a month.

PR #201, opened 2020-09-11
Editorconfig

frigus02:
Add `.editorconfig` file, which hopefully contains how we format files usually. This should help with editor support.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,5 +9,8 @@ insert_final_newline = true
 indent_style = space
 indent_size = 2
 
-[*.sh]
+[{*.sh,Makefile}]
 indent_style = tab
+
+[*.rs]
+indent_size = 4
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,13 @@
+# editorconfig.org
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+indent_style = space
+indent_size = 2
+
+[*.sh]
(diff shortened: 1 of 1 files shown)


alsuren (approved):
Looks good. I might add another PR to make the [tweaks that I have made in my other project]( https://github.com/alsuren/reading-xiaomi-temp/blob/bluez-generated/.editorconfig), but that will cause a bunch of code churn, so let's merge this PR as-is.

frigus02:
To be honest, I would love to switch from spaces to tabs everywhere (except maybe markdown or yaml). The reason is accessibility. Tabs make it easier for everyone to configure their preferred indendation width, which can help with vision impairment.

But yeah. For now I just want to mirror what we have anyway.

I think I need to add something for Rust files and maybe make and docker files 🤔

frigus02:
The shell options are interesting. Are those custom extensions to .editorconfig?

```editorconfig
shell_variant      = posix # like -ln=posix
binary_next_line   = true  # like -bn
switch_case_indent = true  # like -ci
space_redirects    = true  # like -sr
keep_padding       = true  # like -kp
function_next_line = true  # like -fn
```

alsuren (approved):

alsuren:
> The shell options are interesting. Are those custom extensions to .editorconfig?

seems like it. Your editor is unlikely to know how to read them, but shfmt with no args will, from what I've read on https://github.com/mvdan/sh#shfmt and the tickets in that project.

frigus02:
Yep. Sounds fine to me to add those (or some of them).

PR #213, opened 2020-09-11
sync-docker-context to use regustry as build cache

alsuren:
Speed up workspace-service build by using the docker registry as a build cache.

Hrm... This doesn't actually work.

## Numbers

building for db migration check takes ~4mins - we could reduce this time by doing it as part of `make docker-test`?

Fetching the -context image takes about a minute.

For some reason that didn't speed up the build the first time I did it. I wonder if the OSX docker layers I pushed are incompatible with


frigus02:
I think this requires us to use `--cache-from` in our `docker build` commands. Docker doesn't automatically find these cache layers otherwise.

alsuren:
Nope. I give up. I can't get it to skip first `cargo build` even when it manages to successfully pull the image.

frigus02:
Yeah, I tried that before and found it quite hard as well. Or more like: easy to break.

Maybe the best way is to not build in Docker at all and use github actions for caching.

PR #227, opened 2020-09-14
Navigate workspaces

declanslevin:
[Ticket](https://dev.azure.com/futurenhs/FutureNHS/_workitems/edit/2110)

[Figma](https://www.figma.com/file/8gWO0Oz0Htm5UFz9uQcSME/Workspaces?node-id=37%3A0)

## Acceptance Criteria

- [x] Shows list of workspaces a user is member of
- [x] Created reusable WorkspaceDirectoryItem component
- [x] Created theme json in test fixture to pass through in ThemeProvider for tests
- [x] Created horizontal rule under MainHeading

## Pre-review checklist:

- [ ] Tests

- [ ] Documentation

- [ ] Feature flag

## Testing information

- Test plan?

## Test:

- [ ] Tested locally

- Platforms/Browsers:

  - [ ] Internet Explorer 11

  - [ ] Edge

- Accessibility:

  - [ ] Screen Reader

  - [ ] Keyboard Navigation

  - [ ] Magnification

  - [ ] Contrast

  - [ ] Text size 200%

  - [ ] Landscape (Mobile)


frontend/components/MainHeading/MainHeading.tsx
```diff
@@ -1,15 +1,32 @@
 import React from "react";
+
+import styled from "styled-components";
+
+const StyledHorizontalRule = styled.hr`
```
matt-paul, line 5: Naming discussion - i don't really feel we gain much by prefixing components with 'Styled', but harm readability of code and also when the components are viewed within browser console components tree view.  What are your thoughts on that?

frontend/components/Workspaces/WorkspaceDirectoryItem.tsx
```diff
@@ -0,0 +1,74 @@
+import React from "react";
+
+import Link from "next/link";
+import styled from "styled-components";
+
+const StyledImg = styled.img`
+  ${({ theme }) => `
+    display: block;
```
matt-paul, line 8: these can live outside the theme block so that vscode can understand them properly

frontend/components/Workspaces/WorkspaceDirectoryItem.tsx
```diff
@@ -0,0 +1,74 @@
...
+    height: 44px;
+    width: 44px;
+    border-radius: 4px;
+
+    &:hover {
+      &+ div > a > h2 {
```
matt-paul, line 14: this seems quite brittle - is this because of nhs component use?

frontend/components/Workspaces/WorkspaceDirectoryItem.tsx
```diff
@@ -0,0 +1,74 @@
...
+ `}
+`;
+
+interface Props {
+  title: string;
+  id: string;
+}
+
+const WorkspaceDirectoryItem = ({ title, id }: Props) => {
+  return (
```
matt-paul, line 57: Could use implicit return here for better readability?

frontend/pages/admin/workspaces.graphql
```diff
@@ -5,3 +5,18 @@ mutation CreateWorkspaceMutation($title: String!, $description: String!) {
     description
   }
 }
+
+mutation DeleteWorkspaceMutation($id: ID!) {
```
matt-paul, line 9: Don't think this is being used?

frontend/pages/api/graphql.ts
```diff
@@ -8,7 +8,7 @@ const gateway = new ApolloGateway({
   serviceList: [
     {
       name: "workspace-service",
-      url: "http://workspace-service.workspace-service/graphql",
+      url: "http://localhost:3030/graphql",
```
matt-paul, line 11: Not sure this should have changed?

frontend/pages/workspaces/directory.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+    return {
+      props: {
+        workspaces,
+      },
+    };
+  }
+);
+
+const PageContent = styled.section`
```
matt-paul, line 33: Probably time this got put in /components and used from there from now on, i think it's the same as used elsewhere?

frontend/pages/workspaces/directory.tsx
```diff
@@ -0,0 +1,74 @@
...
+  background-color: ${theme.colorNhsukWhite};
+  `}
+`;
+
+interface Props {
+  workspaces: [{ title: string; id: string }];
+}
+
+const WorkspaceDirectory = ({ workspaces }: Props) => {
+  return (
```
matt-paul, line 48: Implicit return?

frontend/components/Workspaces/WorkspaceDirectoryItem.tsx
```diff
@@ -0,0 +1,74 @@
...
+const StyledContainer = styled.div`
+  display: flex;
+  margin-bottom: 40px;
+`;
+
+const StyledTextContainer = styled.div`
+  margin: 0 25px;
+`;
+
+const StyledLink = styled.a`
```
matt-paul, line 31: All this styling seems to be on the H2?

frontend/pages/workspaces/directory.tsx
matt-paul: If you would prefer not to do this in this ticket, then i could do it in my 'view folder in workspace' ticket if that is better for you, let me know

matt-paul:
Do we need cypress / selinium tests in here, or are they coming later?

frontend/components/MainHeading/MainHeading.tsx
```diff
@@ -1,15 +1,32 @@
...
 }
 
 /**
  * This component should be used on every page as the main heading.
  * It's a workaround for a known next.js accessibility issue that doesn't read out
  * headlines on navigation
  */
-const MainHeading = ({ children }: Props) => {
-  return <h1 aria-live="polite">{children}</h1>;
+const MainHeading = ({ children, withBorder }: Props) => {
```
matt-paul, line 23: We might want some kind of test given this component renders differently due to the props.

frontend/components/MainHeading/MainHeading.tsx
trcywu: This is the standard for naming styled-components and to indicate that they have been styled this way. It is consistent across the styled-components docs and also every blog I read in relation to styled-components.

frontend/pages/admin/workspaces.graphql
trcywu: yep will delete

frontend/pages/api/graphql.ts
trcywu: Will change it back. Was this way because mine only worked on localhost

frontend/pages/workspaces/directory.tsx
trcywu: yeah agree, if you can do it in your ticket it'll be great. thanks

frontend/components/Workspaces/WorkspaceDirectoryItem.tsx
declanslevin: It's because the designs asked for the link to be highlighted when the icon is hovered over, which meant I had to traverse the dom tree a little to get to the right selector - as long as the component structure doesn't change this should always work.

frontend/components/MainHeading/MainHeading.tsx
matt-paul: But i think thats more more demonstration purposes isn't it to make it clear? In a 'real project' it feels a bit redundant to me

trcywu: We will be using components that don't get styled and ones that we apply styles ourselves so I think it's relevant to differentiate them.

frontend/components/Workspaces/WorkspaceDirectoryItem.tsx
trcywu: yes, we need to maintain this as an <a> for accessibility purposes and it wraps a h2 so we want the styles to apply to that

frontend/components/Workspaces/WorkspaceDirectoryItem.tsx
matt-paul: yeah, in general we want to be able to change the markup and still have things like the hover functionality work.  You could try doing this with in a more reacty way by using component state perhaps?

trcywu: Could you elaborate, @matt-paul ? Not sure what you're proposing. Why do we need to use component state when we can use css?

matt-paul: Sure - this `&:hover &+ div > a > h2` could easily be accidently broken due to changes in the markup at a later date.  I don't think we should be traversing the dom like this for the above reason, and so suggesting finding a way to avoid this, either by using things that react gives us, or i guess could also be improved by targeting it with an id or class in css too @frigus02 What are your thoughts?

frontend/components/Workspaces/WorkspaceDirectoryItem.tsx
matt-paul: Ah ok. How does styling it in this way help accessability rather than directly styling the h2 with `styled.h2`?

matt-paul: I'm a bit rusty with my frontend so forgive me if i'm missing something!

frontend/pages/workspaces/directory.tsx
```diff
@@ -0,0 +1,73 @@
...
+  );
+};
+
+export default WorkspaceDirectory;
+
+// 1. shell script to post test data via GraphQL
+// 2. Write graphQL `query GetWorkspaces` and then run yarn generate
+// 3. Use graphQL methods in getServerSideProps to fetch data
+// 4. Pass down data as props and print out onto page
+// 5. Check designs with Kim
```
alsuren, line 73: feels like everything on this todo list is done?

frontend/components/Workspaces/WorkspaceDirectoryItem.tsx
frigus02: Hm yeah. It feels a bit weird. I think my biggest concern is probably the exact nesting, which seems like we might break this accidentally. Something like `&+ h2` (or maybe `&+ div h3` if we need the div) would feel okay.

My first thought was to change the style on hover of the `StyledContainer`. Because that way we can target any child `h3` instead of an h3 inside a div, which is next to the img. That changes the logic a bit, though. Not sure if that's okay.

frontend/components/Workspaces/WorkspaceDirectoryItem.tsx
trcywu: It's the same thing. We are directly targeting the h3 that is the child of the <a>. We want to group it together to show the relationship of the h3 being nested in <a>

frontend/components/Workspaces/WorkspaceDirectoryItem.tsx
trcywu: We experimented and `&+ div h3` seems to be the best way forward. Targeting `StyledContainer` results in the whole container area, including white space clickable which is not ideal as someone can accidentally click to the workspace without intending to.

frigus02: Yeah, that was my fear as well. It would look like the entire container is clickable, which is not what we want. I'm happy with `&+ div h3`. It's not perfect, because of the `div`. But it doesn't otherwise rely on the structure too much. 👍

frontend/__tests__/__snapshots__/workspaces-directory.test.tsx.snap
```diff
@@ -0,0 +1,183 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`takes a snapshot of the component 1`] = `
+<DocumentFragment>
+  .c1 {
+  background-color: undefined;
```
trcywu, line 6: sass variables currently come up as undefined @frigus02 @alsuren this is because the sass-loader that we use to create theme does not support jest.

frontend/components/Header/Header.tsx
```diff
@@ -15,11 +15,11 @@ const StyledImg = styled.img`
   display: block;
   height: 40px;
   ${({ theme }) => `
-  @media (max-width: ${theme.mqBreakpoints.tablet}) {
+  @media (max-width: ${theme.mqBreakpoints?.tablet}) {
```
trcywu, line 18: this is the workaround to the sass Jest problem so that i don't get error'd out in my test because they can't find 'tablet of undefined'.

frontend/__tests__/__snapshots__/workspaces-directory.test.tsx.snap
alsuren: https://jestjs.io/docs/en/webpack.html#mocking-css-modules is how I've done it in the past. It's not perfect, but it at least gives you snapshots that break when you change something, and are easy to debug.

frigus02: Sounded like David has a solution for this.

I'm also interested in exploring different ways than snapshot testing. Personally I don't find those snapshots useful. I find them really hard to read and review. But I'm happy to do that at a later point.

alsuren: I will take a look and see if it will work for this project. It might be that we're reaching too deeply into the module for this to work.

trcywu: Have now extracted theme into a fixture and passed through as JSON into tests.

alsuren: @frigus02 It turned out that
a) we're reaching multiple levels deep into the theme object, so identity-object-proxy probably wouldn't work
b) we needed to wrap the component in a `<ThemeProvider/>` in tests (Tracy had code for this from yesterday, but hadn't pushed it)

We ended up extracting the theme as frontend/lib/fixtures/theme.json and using that in tests.

frontend/components/Workspaces/__snapshots__/WorkspaceDirectoryItem.test.tsx.snap
```diff
@@ -0,0 +1,70 @@
...
+
+<div
+    class="c0"
+  >
+    <a
+      class="c1"
+    >
+      <img
+        alt="https://www.nhs.co.uk"
+        src="[object Object]"
```
alsuren, line 62: I think this is being produced by the fileMock thing.  https://jestjs.io/docs/en/webpack#mocking-css-modules suggests some other ways to approach this, but maybe we should do this as a separate PR.

alsuren (approved):

frontend/components/Workspaces/__snapshots__/WorkspaceDirectoryItem.test.tsx.snap
alsuren: ... actually, if you change fileMock.js to `module.exports = "FAKE_WEBPACK_URL";` then it changes all of the snapshots to say `src="FAKE_WEBPACK_URL"`. Maybe that's good enough? 
```
diff --git a/frontend/__mocks__/fileMock.js b/frontend/__mocks__/fileMock.js
index 9cf3c27..076a8fc 100644
--- a/frontend/__mocks__/fileMock.js
+++ b/frontend/__mocks__/fileMock.js
@@ -1 +1 @@
-export default "";
+module.exports = "FAKE_WEBPACK_URL";
```

PR #229, opened 2020-09-15
Publish CreateWorkspace event

frigus02:
[AB#2137](https://dev.azure.com/futurenhs/75407963-f812-43e5-a734-1059bbc036a3/_workitems/edit/2137)

- Backfill CreateWorkspace event

## Pre-review checklist:

- [ ] Tests

## Testing information

- Create a workspace
- Create a subscriber for EventGrid to see if events have been published

## Test

- [ ] Tested locally
- [ ] Tested in cluster


event-models/schema.json
```diff
@@ -59,6 +62,53 @@
...
+                "title": {
+                  "type": "string"
+                },
+                "description": {
+                  "type": "string"
+                }
+              },
+              "required": ["title", "description"]
+            },
+            "output": {
```
alsuren, line 92: (random musings): My instinct says you don't need the output, but I guess you want to record the id that the DB generates.

If we were generating IDs rust-side then we could include that as an "input" and we'd be able to deterministically generate output from input. This would be satisfying, but mostly from a "theoretical purity" perspective (unless you later wanted to take the event stream from a regression and generate test cases from it).

On balance, this pattern of recording inputs and outputs from a mutation function seems simple enough to follow and seems to do what we need it to do (for audit log and reporting purposes).

workspace-service/src/events/mod.rs
```diff
@@ -0,0 +1,23 @@
+use crate::db;
+use anyhow::Result;
+use fnhs_event_models::{CreateWorkspaceEventDataInput, CreateWorkspaceEventDataInputOutput};
+use sqlx::PgPool;
+
+impl From<db::Workspace> for CreateWorkspaceEventDataInputOutput {
```
alsuren, line 6: (maybe nothing you can do about this): Where does the name `CreateWorkspaceEventDataInputOutput` come from? It appears to describe .data.output. Why does it have Input in the name? :/

workspace-service/src/graphql/workspaces.rs
```diff
@@ -71,10 +95,20 @@ impl WorkspacesMutation {
         context: &Context<'_>,
         workspace: NewWorkspace,
     ) -> FieldResult<Workspace> {
-        // TODO: Add event
         let pool = context.data()?;
-        let workspace = db::Workspace::create(workspace.title, workspace.description, pool).await?;
-        Ok(workspace.into())
+        let event_client: &EventClient = context.data()?;
+
+        let input: CreateWorkspaceEventDataInput = workspace.into();
```
alsuren, line 101: (this came out sounding more sure of myself than I really am): I feel like the NewWorkspace graphql type already satisfies many of the properties you want from your input object. In my mind, NewWorkspace should be thought of as part of our domain model, and CreateWorkspaceEventDataInput should be thought of as an implementation detail of how we do serialisation.

I would therefore make create_workspace() accept &NewWorkspace, and I would delay calling NewWorkspace::into() until the time when I call publish_events().

If I wanted to make sure that I wasn't missing anything from the event, I would impl from in the other direction, and write a test that checks that the round-trip is fine.

I have not thought about how I would deal with things that we fish out of context (like the user id). Maybe we say `input: NewWorkspace, context: FieldsPluckedFromContext, output: Workspace`. I guess in your model, we pluck everything we need out of Context?

workspace-service/src/events/mod.rs
frigus02: Exactly my thoughts. This is generated by `schemafy`. I have no idea why it adds "Input" to that name. Might be a bug.

frigus02:
I just had a chat with Stu about events. He had an idea, which I think might work quite nicely.

Our thought until now was to generate an event for all commands in the system (e.g. workspace create, file edited, folder deleted). Those events would contain all data necessary, basically so they could be replayed. Replaying events is probably mostly theoretical, but building for it seems nice because it means every detail has to be in the event. In the end this would lead to events, which can easily be consumed by new applications and services.

Stu's idea was instead to generate specific events for each consumer. This means the workspace service would generate a "folder created" event for analytics and another one for the audit log (and possiby more events if we have more consumers).

What's the benefit?

- Events are tailored for the consumer. This makes it easy to design them. You typically know exactly what the consumer needs.
- Events should change less often. Changes to events sound painful in the first approach when we have to update all consumers or publish both event versions simultaniously.

What are the downsides?

- When you add a new consumer, you have to add more events to each service that the consumer is interested in.

event-models/schema.json
frigus02: I quite like the idea of generating the ID in Rust instead of Postgres. It sounds like it could make the code easier to read. I might give this a go.

alsuren:
I agree. The pseudo-event-sourcing model is mostly nice for theoretical reasons. I'm happy to take a step back and just emit the events that we know we need.

> Events should change less often. Changes to events sound painful in the first approach when we have to update all consumers or publish both event versions simultaniously.

The dance for updating events probably doesn't involve publishing multiple events. I've always seen it look more like:
* write new v2 event model definition and make all consumers capable of consuming both v1 and v2 versions of the event
* make the producer switch from publishing v1 to publishing v2
* kill off v1 event model definition and all code that consumes v1 model

StuartHarris:
There's a section in [this post](https://dev.to/olibutzki/why-event-sourcing-is-a-microservice-anti-pattern-3mcj) that explains the Customer/Supplier concept from DDD that Jan was refering to. Not married to it but it might help. Let's chat together about it when we get a moment.

matt-paul:
> There's a section in [this post](https://dev.to/olibutzki/why-event-sourcing-is-a-microservice-anti-pattern-3mcj) that explains the Customer/Supplier concept from DDD that Jan was refering to. Not married to it but it might help. Let's chat together about it when we get a moment.

I'm interested to hear about this if you haven't already discussed...

event-models/schema.json
frigus02: We now explicitely decided to only use events for domain events. We might also make a separate event per consumer. This means the event doesn't need to contain all data, only things relevant to the consumer. This means the ID generation is not a problem we need to think about now :)

workspace-service/src/events/mod.rs
frigus02: We flattened the schema for now, which gets rid of this problem.

Long term it might be nicer to generate the Rust types in another way that results in actual Rust files, which we can commit to the repo. This would make reading those types much easier.

workspace-service/src/graphql/workspaces.rs
frigus02: I guess that's also mostly addressed by the consumer specific events.

workspace-service/src/lib.rs
```diff
@@ -111,4 +120,46 @@ mod tests {
 
         Ok(())
     }
+
+    #[async_std::test]
+    async fn creating_workspace_emits_an_event() -> anyhow::Result<()> {
```
alsuren, line 125: we do not like this. This is outside-in.
- [ ] Move the other tests in this module out into /test style integration tests, 
- [ ] split graphql resolvers from business logic so that business logic can be unit tested.

alsuren: we ended up deleting the pseudo-integration tests, and we can talk about the correct approach this afternoon.

We split the resovler into a separate function to facilitate testing, but kept it in the same file.

edcompton (approved):
🔥 🔥 🔥 🔥 🔥

StuartHarris (approved):
This looks great! (I think the deleted tests aren't very useful anyway, so I don't think we need to worry about them later). 
I think we should have a conversation at some point about what we do if, for example, workspace creation succeeds but the event publish fails,. But good for now.

PR #230, opened 2020-09-15
AB#1789 - Show Folders in Workspace (UI)

matt-paul:
[Ticket](https://dev.azure.com/futurenhs/FutureNHS/_workitems/edit/1789)

[Figma](https://www.figma.com/file/tYFsqhG61CleWS1tUapAar/Source-of-Truth-UI?node-id=47%3A0)

[Deployed Here]https://fnhs-dev-david.westeurope.cloudapp.azure.com/workspaces/2843ac55-9325-4785-97be-f8140533fc72/folders/8d45ef39-8a8a-4bed-be37-a2ae5d389c08)

Screenshot:
Screenshot 2020-09-21 at 17 01 17

Folders list closed:
Screenshot 2020-09-21 at 17 20 22

Folder selected in navigation, folder homepage visible:
Screenshot 2020-09-21 at 17 20 09

As a user
I want to see folders in workspaces
So that I can navigate to the file I want
## Pre-review checklist:

- [x] Tests (will do Selenium tests in a separate PR)

- [x] Documentation

- [ ] Analytics (will happen in a separate PR)

## Testing information

1) Navigate to a workspace homepage
2) Admire nav bar to the left of screen
3) Note list of folders in the nav bar (use `create-folders`script in `workspace-service/scripts` to create if you don't have any in workspace already)
4) Verify that when you select a folder, the folder icon changes to 'open', your url changes to the 'folder homepage', and that folder is highlighted'

## Test:

- [ ] Tested locally

- Platforms/Browsers:

  - [ ] Internet Explorer 11

  - [ ] Edge

- Accessibility:

  - [ ] Screen Reader

  - [ ] Keyboard Navigation

  - [ ] Magnification

  - [ ] Contrast

  - [ ] Text size 200%

  - [ ] Landscape (Mobile)


frontend/components/NavListItem/NavListItem.tsx
```diff
@@ -0,0 +1,51 @@
...
+`;
+
+const icons: { [key: string]: { [key: string]: string } } = {
+  folder: {
+    closed: require("../../public/closedFolder.svg"),
+    open: require("../../public/openFolder.svg"),
+  },
+};
+
+const NavListItem = ({ active, item, icon, workspaceId }: Props) => (
```
alsuren, line 40: let's remove "icon" prop here, and think about it properly when we have another item type to implement.

frontend/components/NavSection/NavSection.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+  const openChevron = require("../../public/chevronOpen.svg");
+  const closedChevron = require("../../public/chevronClosed.svg");
+  return (
+    <Section>
+      <div>
+        <h4>{title}</h4>
+        <img
+          src={open ? openChevron : closedChevron}
+          onClick={() => setOpen(!open)}
```
frigus02, line 38: Minor thing. Since this is clickable, we should use the correct semantic markup. I think in this case the image should be in a button.

frigus02: Maybe you had this on your list anyway. I just saw this in the demo, so thought I mention it 🙂

frontend/components/Navigation/Navigation.tsx
```diff
@@ -28,11 +29,22 @@ const Nav = styled.nav`
 
 const Header = styled.header`
   padding-bottom: 20px;
+  h3 {
```
matt-paul, line 32: Think we can remove this

workspace-service/scripts/create-folder-if-needed.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-set -euo pipefail
+set -exuo pipefail
```
alsuren, line 3:
```suggestion
set -euo pipefail
```

frontend/components/NavListItem/NavListItem.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+type ListItem = Pick<Folder, "id" | "title">;
+
+interface Props {
+  active: boolean;
+  item: ListItem;
+  workspaceId: string;
+}
+
+const ListItem = styled.li<any>`
```
trcywu, line 16: Why are we not naming the components we create and style with the prefix “Styled”? This was discussed before. We need it to differentiate between our own styled elements, components we import from the NHS react component library, and HTML elements. It’s important to differentiate. The “Styled” prefix is the correct styled components convention. It is used in all **real** projects. We should be consistent.

trcywu: For example: We need to differentiate html <button> vs imported component from nhs <Button> and our own styled version <StyledButton>

frontend/components/NavListItem/NavListItem.tsx
```diff
@@ -0,0 +1,65 @@
...
+  }
+`;
+
+const icons: { [key: string]: string } = {
+  closed: require("../../public/folderClosed.svg"),
+  open: require("../../public/folderOpen.svg"),
+};
+
+const NavListItem = ({ active, item, workspaceId }: Props) => (
+  <ListItem style={active ? { background: "#ffeb3b", fontWeight: 700 } : {}}>
```
trcywu, line 55: If we used `StyledListItem` here then we can clearly differentiate, which elements here were styled, which are not and which are imported.

frontend/components/NavListItem/NavListItem.tsx
matt-paul: I'm not sure there was a conclusion to the exact naming convention we were going to use going forward, simply a difference of opinions at this point?  David made a [suggestion](https://github.com/FutureNHS/futurenhs-platform/pull/253#discussion_r492199324) in Declans PR to live with the two for a bit, and see how things settle once we have a number of components in place, as well as possible solution combining the two methods.  I think this is about finding a naming convention that the team is happy with in the longer term, and that addresses concerns raised previously on prefixing everything with 'Styled'. (if you look at the docs for styled components you will notice that they don't do that themselves, and i know from debate on issue that there are multiple conventions on this for real projects out there). I think we need to chat as a team at a suitable point to settle on something that reflects the tradeoffs of both.

frontend/components/NavListItem/NavListItem.tsx
matt-paul: In this instance i don't see value in that when the component is small and its easy to see that information elsewhere in the file, whilst making harder to grok. (we never would have called a css class .styledNavItem or .navItemWithStyles or something to differentiate between classes that are used for css against those being used by js). Again, would suggest we defer to David's idea (linked in above comment) to live with both for a bit and settle on what we like as a team together and ideally find a system that gives us the benefits of both approaches, perhaps when Jan is back next week?

frontend/components/NavListItem/NavListItem.tsx
matt-paul: Resolving this after discussion / agreement on approach in slack.

frontend/components/NavListItem/NavListItem.tsx
matt-paul: As above, resolving after discussion in slack

frontend/components/Navigation/Navigation.tsx
```diff
@@ -0,0 +1,88 @@
...
+import Link from "next/link";
+import styled from "styled-components";
+import { v4 as uuid } from "uuid";
+
+import { Folder, Workspace } from "../../lib/generated/graphql";
+import { NavListItem } from "../NavListItem";
+import { NavSection } from "../NavSection";
+
+const Nav = styled.nav`
+  padding-top: 24px;
```
edcompton, line 12: any discussion about separating styles into a separate `styles.tsx` or similar to sit alongside the component in the directory?

frontend/pages/workspaces/[id]/folders/[folderId].tsx
```diff
@@ -0,0 +1,81 @@
...
+import { getSdk, Folder, Workspace } from "../../../../lib/generated/graphql";
+
+export const getServerSideProps: GetServerSideProps<Props> = requireAuthentication(
+  async (context) => {
+    const client = new GraphQLClient(
+      "http://workspace-service.workspace-service/graphql"
+    );
+    const sdk = getSdk(client);
+    const folderId = (context.params?.folderId as string) || "";
+    const workspaceId = (context.params?.id as string) || "";
```
Comment (edcompton) [inline frontend/pages/workspaces/[id]/folders/[folderId].tsx:22]: Should this be more explicit at the point it's attached to `context`? e.g. `workspaceId` or similar?

workspace-service/scripts/create-default-test-folders.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
```
edcompton, line 1: one question for the future, rather than for this script - is it better to do this sort of action with a Makefile?

edcompton (commented):
Looks great in general 😄 excitingly for me, lots to understand in there! Couple of small questions/suggestions that are probably due to a lack of context. 

One more general point - quite a big PR at 1500 lines (appreciate snapshots make up a fair stack of that), but in terms of review may be easier to break the work up further in future. Don't know what the vibe is in terms of what's been discussed in breaking up tickets, so may just require clarification.

frontend/components/NavListItem/NavListItem.tsx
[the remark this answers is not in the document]
matt-paul: Good spot thanks, that's an oversight

frontend/components/NavSection/NavSection.tsx
```diff
@@ -0,0 +1,53 @@
...
+    justify-content: space-between;
+    align-items: start;
+  }
+  button {
+    border: none;
+    background: inherit;
+  }
+  ${({ theme }) => `
+    h4 {
+      color: ${theme.colorNhsukBlue};
```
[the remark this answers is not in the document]
matt-paul: 👍

frontend/components/NavListItem/index.tsx
```diff
@@ -0,0 +1,3 @@
+import NavListItem from "./NavListItem";
+
+export { NavListItem };
```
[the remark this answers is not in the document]
matt-paul: It's not a stupid question - personally i would like us to review the need for this on every component as i don't have a good answer.

frontend/components/NavSection/index.tsx
```diff
@@ -0,0 +1,3 @@
+import NavSection from "./NavSection";
+
+export { NavSection };
```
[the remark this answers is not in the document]
matt-paul: As above 🙂

frontend/components/Navigation/Navigation.tsx
matt-paul: No discussion yet.   Personally i quite like having the styles in the same file, provided it doesn't get unwieldy. I think the apprach we have taken is definitely open for further discussion though, perhaps worth trying in a following PR for comparison?

frontend/pages/workspaces/[id]/folders/[folderId].tsx
Comment (matt-paul) [reply frontend/pages/workspaces/[id]/folders/[folderId].tsx:22]: That's not a bad point.  The 'id' was created before folders were a thing, so made more sense then.  Will look to change this in this PR provided there isn't too much of a cascade of consequences! (should be fine but might double check with David)

workspace-service/scripts/create-default-test-folders.sh
matt-paul: Will leave this question for @alsuren as he's better positioned to answer this.

matt-paul:
> Looks great in general 😄 excitingly for me, lots to understand in there! Couple of small questions/suggestions that are probably due to a lack of context.
> 
> One more general point - quite a big PR at 1500 lines (appreciate snapshots make up a fair stack of that), but in terms of review may be easier to break the work up further in future. Don't know what the vibe is in terms of what's been discussed in breaking up tickets, so may just require clarification.

Yeah, definitely been fighting to reduce the size of PR's.  We have split out the e2e tests into a follow up PR, but still room for improvement on this i think.  If i was doing this again i probably would have split the folder homepage and empty nav into one PR, and the folders inside into another, i think that might have worked better

frontend/pages/workspaces/[id]/folders/[folderId].tsx
Comment (matt-paul) [reply frontend/pages/workspaces/[id]/folders/[folderId].tsx]: Had a little look at this to remember whats going on.. it seems it's just a simple enough renaming of a file and folder within `pages/workspaces` and then Next performs its magic. Seems easy enough to change (along with the related __tests__ folder structure to reflect new naming.  What are you thoughts @alsuren ? ps changes in latest commit

workspace-service/scripts/create-default-test-folders.sh
alsuren: Stu made the point during your cluster setup that using makefiles forces you to keep things simple. I suppose by not forcing ourselves to put everything in makefiles from the start, we have allowed ourselves some get-out-of-jail-free things like the ENVIRONMENT argument.

There is also the argument for "doing it properly" and writing the example-data-population scripts in the same language that we will write our kahootz migration scripts in (or in the ideal case, just have a cassette of canned kahootz data in blob storage somewhere, and run the migration scripts on it whenever you want a new cluster). This would require us to write kahootz migration scripts in lock-step with our features though, which would cause quite a lot of context switching and probably slow us down.

If I was being forward-thinking, I suppose bash is the worst choice for this, but it's a quick fix that gets us out of a bind, so I went with it.

frontend/pages/workspaces/[id]/folders/[folderId].tsx
Comment (alsuren) [reply frontend/pages/workspaces/[id]/folders/[folderId].tsx]: looks good.

frontend/components/Navigation/Navigation.tsx
alsuren: I have found that having the whole component in one file a bit easier to navigate. That's the goal css-in-js and react-hooks are striving for, I think. At some point there was a suggestion that we could use CSSModules, but we decided that we like styled-components too much.

frontend/components/NavListItem/index.tsx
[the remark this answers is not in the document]
edcompton: fair enough, discussion to be had outside of this PR

frontend/components/Navigation/Navigation.tsx
edcompton: Cool, sounds good - as long as you keep your components nice and small I can see that it works well.

frontend/components/NavListItem/index.tsx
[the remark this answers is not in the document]
alsuren: I started this PR with no opinions either way. It has already shown the useful property that the snapshots live next to the components and tests in the github PR. It also makes our component tree more fractal, which I quite like, and the cost is low.

I say we keep going with this for a bit longer, and see if we all grow to like it. I don't think that the refactor will be that difficult if we decide to flatten things. I have even seen autoformatters that know how to add/remove these levels of nesting.

workspace-service/scripts/create-default-test-folders.sh
edcompton: fair play. Obviously not suggesting a re-write at this point, more a question of value for future examples of a similar nature.

trcywu (approved):

frontend/components/NavListItem/index.tsx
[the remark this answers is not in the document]
edcompton: I guess what I'm unclear on is the need for a separate `NavListItem.tsx` - could that content just be included in the `index.tsx`? Couldn't the snapshots and tests live next to the `index.tsx` without the need for the `NavListItem.tsx`? I guess I'm a bit unclear as to what problem this approach solves, but may be missing something obvious. As above, happy to clarify outside of this PR 👍

edcompton (approved):

PR #231, opened 2020-09-16
Speed up Rust builds: build outside Docker and caching cargo dirs

frigus02:
I tested ways of speeding up Rust builds:

- build outside Docker and caching cargo dirs (this one)
- Docker layer caching #232

Caching Docker layers is nice, because it means you can replicate the same Docker build locally. But caching is slow. It takes 7mins to restore and 7mins to store the cache (16mins total build time when everything is cached).

Building outside Docker and caching the cargo directories is much faster (3mins total build time when everything is cached). Downside: you cannot create a Docker image locally anymore. Because we build on Mac the binary doesn’t run on Linux. But in CI it's okay because the build runs on Ubuntu.

I think I'm in favor if this one here. I think it's super rare that we need to replicate local Docker builds and this one makes the build much quicker.

### DIFF
--- a/.github/workflows/workspace-service-branch.yaml
+++ b/.github/workflows/workspace-service-branch.yaml
@@ -36,6 +36,7 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           toolchain: stable
+          components: clippy
       - name: Cache sqlx
         id: cache-sqlx
         uses: actions/cache@v2
--- a/.github/workflows/workspace-service-branch.yaml
+++ b/.github/workflows/workspace-service-branch.yaml
@@ -34,7 +34,6 @@ jobs:
           username: ${{ secrets.REGISTRY_USERNAME }}
           password: ${{ secrets.REGISTRY_PASSWORD }}
       - uses: actions-rs/toolchain@v1
-        working-directory: futurenhs-platform
       - name: Cache sqlx
         id: cache-sqlx
         uses: actions/cache@v2
--- a/.github/workflows/workspace-service-branch.yaml
+++ b/.github/workflows/workspace-service-branch.yaml
@@ -36,6 +36,7 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           toolchain: stable
+          override: true
           components: clippy
       - name: Cache sqlx
         id: cache-sqlx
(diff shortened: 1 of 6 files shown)


workspace-service/Dockerfile
```diff
@@ -34,28 +8,7 @@ RUN apt-get update && apt-get install -y \
...
-ARG TEST_DATABASE_URL=postgres://postgres:postgres@172.17.0.1:5432
-ENV TEST_DATABASE_URL=$TEST_DATABASE_URL
-
-RUN cargo test --release
-
-# Create the release
-FROM release-context AS release
-
-COPY --from=build /usr/src/workspace-service/target/release/workspace_service /
+COPY target/release/workspace_service /
```
alsuren, line 11: I think that this approach could work if we add a selfcheck mode to the binary, and then run it like:

`RUN env SELFCHECK_ONLY=1 /workspace_service`

If the binary is compiled correctly and runnable then the docker build will be successful, and if not then it will explode with an error telling you that it is intended to be built via github actions.

we could also do a `uname -a` check in `make docker-build` but that would be less bullet-proof.

frigus02: Oh nice. Yeah, I like that idea.

I recently forgot to install libcurl, which meant the Docker image built successfully but out binary failed to run. Does it test that, too? 🤔

alsuren: > Does it test that, too?

I don't know. You would have to test it ;-)

There's a reasonable chance that it would also catch this case, as long as our libraries are not doing any crazy tricks to lazily-load *.so libraries.

frigus02: Yep, that's what I'm hoping. The simple test works. I did `cargo build --release` and `make docker-build` on my Mac and the Docker build fails 🎉

frigus02: Sadly it did not catch a missing libcurl 😞

.github/workflows/workspace-service-branch.yaml
```diff
@@ -36,27 +36,45 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           toolchain: stable
+      - name: Install clippy
+        working-directory: futurenhs-platform
+        run: rustup component add clippy
```
frigus02, line 41: We should be able to add components using the `actions-rs/toolchain@v1` step. But it seems it doesn't work together with our `rust-toolchain` file for some reason.

frigus02: Fixed together with David 🎉

PR #232, opened 2020-09-17
Speed up Rust builds: Docker layer caching

frigus02:
The action https://github.com/satackey/action-docker-layer-caching should do most of the work for us.


frigus02:
Hm. That doesn't seem to work. I restarted the last build after it was successful. But it doesn't seem to make any use of the cached/restored layers.

frigus02:
Restoring and storeing the cache takes quite a bit of time. Let's see how this is on a second run...

image

frigus02:
Hm. Fast build now but slow caching

image

frigus02:
Replaced by #231

PR #236, opened 2020-09-18
Docker Standards

frigus02:
For the Architecture Review Board call we would like to have some short form standards on Docker & Kubernetes.

These seem like the basic things to look for with Docker and Kubernetes. I don't think there is anything controversial in there. I think this is probably good to merge. We can always extend it later on.

### DIFF
--- a/docs/tech-stack/docker-kubernetes.md
+++ b/docs/tech-stack/docker-kubernetes.md
@@ -1,0 +1,20 @@
+# Docker & Kubernetes
+
+## Docker
+
+- Keep base images up to date (dependabot)
+- Use official and tested base images where possible (debian)
+- Use small images without unnecessary software packages, which can add security vulnerabilities (debian slim)
+- Scan images (Azure Security Center)
+- Don't store secrets in images. Instead pass then in at runtime using environment variables
+- Don't run as root (TODO)
+- Drop unnecessary capabilities (TODO)
+- Define resource limits (TODO)
+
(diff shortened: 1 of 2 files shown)


docs/tech-stack/docker-kubernetes.md
```diff
@@ -0,0 +1,20 @@
...
+- Don't run as root (TODO)
+- Drop unnecessary capabilities (TODO)
+- Define resource limits (TODO)
+
+## Kubernetes
+
+- Use managed cluster (e.g. AKS, GKE, EKS), which secures access to nodes and master
+- Use RBAC
+- Use namespaces to give pods minimal access to secrets
+- Enforce network policies, e.g. with NetworkPolicy or Istio (TODO)
```
StuartHarris, line 19: These are great! During a pen test at MHRA, they suggested a "security in depth" approach of using both network security policies (k8s) and authorization policies (Istio), which we are in the process of doing

frigus02: Nice. I should probably come back to this PR and tidy up the notes a bit. I'm still hoping Linkerd introduces policies based on mTLS before we go live and we can make use of that 🤞 🙂

matt-paul (approved):
Looks good!

StuartHarris (approved):
Looks good to me. We should schedule some time to review our current K8s RBAC and Istio RBAC, etc

PR #253, opened 2020-09-21
Navigate my workspaces - "View My Workspaces" link

declanslevin:
[Ticket](https://dev.azure.com/futurenhs/FutureNHS/_workitems/edit/2111)

[Figma](https://www.figma.com/file/tYFsqhG61CleWS1tUapAar/Source-of-Truth-UI?node-id=57%3A28)

## Acceptance Criteria

- [x] Use header component as a basis: https://service-manual.nhs.uk/design-system/components/header

- On desktop 
  - [x] Top bar implemented with the "View my workspaces" button
  - [x] When clicked, goes to My Workspaces page
  - [x] Tob bar contains menu button which displays dropdown menu when clicked, over content and aligned to the right
  - [x] "My workspaces" dropdown menu list item is hidden on desktop view

- On mobile 
  - [x] Menu button shows "My workspaces" link, which transitions to My Workspaces page

## Pre-review checklist:

- [x] Tests

- [ ] Documentation

- [ ] Feature flag

## Testing information

- Test plan?

## Test:

- [ ] Tested locally

- Platforms/Browsers:

  - [ ] Internet Explorer 11

  - [ ] Edge

- Accessibility:

  - [ ] Screen Reader

  - [ ] Keyboard Navigation

  - [ ] Magnification

  - [ ] Contrast

  - [ ] Text size 200%

  - [ ] Landscape (Mobile)


frontend/components/Header/NavHeader.tsx
```diff
@@ -0,0 +1,152 @@
...
...0.86676 10 0.86676C10.3294 0.86676 10.6506 0.931543 10.9545 1.05931L19.7597 4.76034C19.9827 4.85406 20 5.06192 20 5.1234C20 5.18487 19.9827 5.39274 19.7597 5.48644L10.9545 9.18748C10.6506 9.31524 10.3294 9.38002 10 9.38002C9.67059 9.38002 9.34944 9.31524 9.04547 9.18748L0.240284 5.48644ZM19.7597 11.2471C19.9827 11.3408 20 11.5487 20 11.6101C20 11.6716 19.9827 11.8795 19.7597 11.9732L10.9545 15.6742C10.6506 15.802 10.3294 15.8668 10 15.8668C9.67059 15.8668 9.34944 15.802 9.04547 15.6742L0.240284 11.9732C0.0173271 11.8795 0 11.6716 0 11.6101C0 11.5487 0.0173271 11.3408 0.240284 11.2471L2.5921 10.2586L8.85456 12.8908C9.22103 13.0448 9.61051 13.1219 10 13.1219C10.3895 13.1219 10.779 13.0448 11.1454 12.8908L17.4079 10.2586L19.7597 11.2471ZM19.7597 8.72981L10.9545 12.4308C10.6506 12.5586 10.3294 12.6234 10 12.6234C9.67059 12.6234 9.34944 12.5586 9.04547 12.4308L0.240284 8.72981C0.0173271 8.63611 0 8.42824 0 8.36677C0 8.30529 0.0173271 8.09743 0.240284 8.00371L2.5921 7.01518L8.85456 9.64745C9.22103 9.80148 9.61051 9.8785 10 9.8785C10.3895 9.8785 10.779 9.80148 11.1454 9.64745L17.4079 7.01518L19.7597 8.00371C19.9827 8.09743 20 8.30529 20 8.36677C20 8.42824 19.9827 8.63611 19.7597 8.72981Z"
```
matt-paul, line 110: Does this need to be inline? 🤔

frontend/components/Header/NavHeader.tsx
```diff
@@ -0,0 +1,152 @@
...
+const NavHeader = () => {
+  return (
+    <StyledHeader>
+      <StyledHeaderContainer>
+        <FnhsLogo />
+        <StyledHeaderLogo href="https://www.nhs.uk" />
+        <Header.MenuToggle />
+      </StyledHeaderContainer>
+      <StyledHeaderNav>
+        <Link href="/workspaces/directory">
```
alsuren, line 104: when I checked this out locally, I noticed that when you mouse-over, the mouse cursor changes to a caret rather than a finger (might not be the correct technical terms). In our other PR, we had a styled.a and it wasn't rendering correctly. We ended up setting passHref (as suggested in https://github.com/vercel/next.js/issues/1942#issuecomment-437680021). I'm not sure if this will work for you though, so be sure to check for yourself whether it fixes the cursor.

matt-paul:
I'm not sure i understand why some of the svgs don't live in the public folder with the others- perhaps there is a reason that I haven't understood yet?

frontend/components/Header/NavHeader.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+import { FnhsLogo } from "../Svg";
+
+const StyledHeader = styled(Header)`
+  ${({ theme }) => `
+    background-color: ${theme.colorNhsukWhite};
+  `}
+`;
+
+const StyledHeaderContainer = styled.div`
```
alsuren, line 15: I wasn't sure what the StyledWhatever naming scheme was about in the previous PR, but I'm a bit new to styled components and I hadn't appreciated that there is a `styled(Header)...` construct that's different from the `styled.div...` construct.

I'm wondering if we should keep the StyledWhatever naming scheme for places where we're doing `const StyledWhatever = styled(Whatever)...` and use simpler names when we're doing `const HeaderContainer = styled.div`?

It may be that the correct answer is to leave the names as they are for a bit, until we've had a few rounds of everyone reviewing each other's code, and agree on a naming convention later and rename everything in one go. (might be worth you taking a quick review of https://github.com/FutureNHS/futurenhs-platform/pull/230 as well, and deciding what you like/dislike)

frontend/pages/workspaces/[id].tsx
```diff
@@ -58,7 +58,7 @@ const WorkspaceHomepage = ({ workspace }: Props) => (
   <>
     <Head title={workspace.title} />
     <PageLayout>
-      <Header />
+      <NavHeader />
```
Comment (alsuren) [inline frontend/pages/workspaces/[id].tsx:61]: It looks like this causes some snapshots to change.

If you don't mind your laptop trying to take off, I would recommend running `yarn test -u --watchAll` in a terminal tab (like how you run `yarn dev`). This will help you to commit your snapshot changes at the same time as your code.

You also have a warning that says 
```
Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?

Check the render method of `NavHeader__StyledHeaderNavItem`.
```
I've not looked into it any further. Can look at it with you tomorrow afternoon if you want.

frontend/components/Header/NavHeader.tsx
```diff
@@ -0,0 +1,152 @@
...
+  return (
+    <StyledHeader>
+      <StyledHeaderContainer>
+        <FnhsLogo />
+        <StyledHeaderLogo href="https://www.nhs.uk" />
+        <Header.MenuToggle />
+      </StyledHeaderContainer>
+      <StyledHeaderNav>
+        <Link href="/workspaces/directory">
+          <StyledHeaderNavItem>
```
matt-paul, line 105: Viewing on desktop, it feels weird semantically for this single link to be in a list as a list element.  I think it might be worth treating the list for mobile, and the 'my workspaces' link as two completely separate things.

frontend/components/Header/NavHeader.tsx
```diff
@@ -0,0 +1,152 @@
...
+        </StyledHeaderNavItem>
+        <StyledHeaderNavItem id="nav-item-hide" href="#">
+          <StyledIcon src={require("../../public/icons/profile.svg")} alt="" />
+          View profile
+        </StyledHeaderNavItem>
+        <StyledHeaderNavItem id="nav-item-hide" href="#">
+          <StyledIcon src={require("../../public/icons/help.svg")} alt="" />
+          Help
+        </StyledHeaderNavItem>
+        <StyledHeaderNavItem id="nav-item-hide" href="#">
```
matt-paul, line 143: there should never be multiple elements on a page with the same id

frontend/public/icons/dashboard.svg
```diff
@@ -0,0 +1,3 @@
+<svg width="18" height="19" viewBox="0 0 18 19" fill="none" xmlns="http://www.w3.org/2000/svg">
```
matt-paul, line 1: Check the sizes here with Kim and also check your export from Figma - would normally expect the icons in a list to all be the same size (eg 24 x 24). This allows design to align the group in a way they are happy with whilst making it easy for you to just slot them all in and line up in the expected way

frontend/components/Header/NavHeader.tsx
trcywu: `Styled` should prefix whatever we have applied our own styling too.

frontend/components/Header/NavHeader.tsx
declanslevin: It was tricky targeting the .svg files to change the "fill" value when the background colour changes. 
The solution I went with was to use a "styled" svg component, which requires the paths inline in its current implementation. 
To remove the inline paths it would be necessary to create a styled svg component for each icon - Which I can do though I _think_ it might introduce a problem that you can't add further styling after importing (need to investigate this more though).

frontend/components/Header/NavHeader.tsx
declanslevin: I just hid the other links in the list for now to match the design I was working from.

frontend/components/Header/NavHeader.tsx
declanslevin: I'm aware of this, I did put it in a comment above the first occurrence of the id to say it was a temporary thing to hide the unnecessary list items for now. I was trying to target them with a class but the specificity wasn't strong enough for what ever reason so just used an id with the knowledge that it would be removed later.

frontend/components/Header/NavHeader.tsx
matt-paul: So i thiiiiink an alternative would be just to use two different svgs (they are super small anyway) pulled in with  tags, and just show one or the other.  This would allow us to not worry about needing to change values such as fill ourselves (only really needed if doing animations which i don't think we are using here?)  Would that work?

matt-paul: So it would be something like this (from the nav list item)
```
const icons: { [key: string]: string } = {
  closed: require("../../public/folderClosed.svg"),
  open: require("../../public/folderOpen.svg"),
};

const NavListItem = ({ active, item, workspaceId }: Props) => (
  <ListItem style={active ? { background: "#ffeb3b", fontWeight: 700 } : {}}>
    <Link href={`/workspaces/${workspaceId}/folders/${item.id}`}>
      <a>
        <img src={active ? icons["open"] : icons["closed"]} />

etc

```

frontend/components/Header/NavHeader.tsx
declanslevin: It did work 👍

frontend/components/Header/NavHeader.tsx
```diff
@@ -0,0 +1,200 @@
...
+    }
+    return icon;
+  };
+
+  return (
+    <Link href={link} passHref>
+      <StyledHeaderNavItem
+        onMouseOver={toggle}
+        onMouseOut={toggle}
+        onMouseDown={toggleActive}
```
alsuren, line 163: Hrm. I hadn't realised that this was the alternative to inlined svgs. :/

frontend/pages/workspaces/directory.tsx
```diff
@@ -48,7 +48,7 @@ interface Props {
 const WorkspaceDirectory = ({ workspaces }: Props) => {
   return (
     <PageLayout>
-      <Header />
+      <NavHeader />
```
alsuren, line 51: rename Header to LoginHeader?

frontend/components/Header/index.tsx
```diff
@@ -1,3 +1,15 @@
...
+import { NavMenuButton } from "./Nav/NavMenuButton";
+import NavMenuListItem from "./Nav/NavMenuListItem";
+import NavHeader from "./NavHeader";
 
-export { Header };
+export {
+  Header,
+  NavHeader,
+  NavList,
+  NavMenuListItem,
```
alsuren, line 12: (not sure what the answer is) make a `Nav/index.ts` for these or just don't bother reexporting them at all?

alsuren: maybe we can talk to Tracy about component folder structures tomorrow.

alsuren:
- [ ] snapshot tests

frontend/components/Header/index.tsx
alsuren: Finally got around to talking to Tracy. She pointed out that NavHeader and Nav/* probably doesn't want to live inside Header.

Given that we were toying with the idea of renaming Header to LoginHeader anyway https://github.com/FutureNHS/futurenhs-platform/pull/253#discussion_r496101911
this would change it from:
```
frontend/components/Header/Header.tsx
frontend/components/Header/Nav/NavList.tsx
frontend/components/Header/Nav/NavMenuButton.tsx
frontend/components/Header/Nav/NavMenuListItem.tsx
frontend/components/Header/Nav/index.tsx
frontend/components/Header/NavHeader.tsx
frontend/components/Header/index.tsx
```
to:
```
frontend/components/LoginHeader/LoginHeader.tsx
frontend/components/LoginHeader/index.tsx
frontend/components/NavHeader/NavList.tsx
frontend/components/NavHeader/NavMenuButton.tsx
frontend/components/NavHeader/NavMenuListItem.tsx
frontend/components/NavHeader/NavHeader.tsx
frontend/components/NavHeader/index.tsx
```

frontend/components/Header/Nav/NavList.tsx
```diff
@@ -79,7 +103,9 @@ const NavList = ({ navItems, setMenuOpen }: NavListProps) => {
     <StyledNavMenuContainer>
       <p>
         <span>Menu</span>
-        <StyledCloseIcon onClick={() => setMenuOpen(false)} />
+        <StyledCloseIconWrapper>
+          <StyledCloseIcon onClick={() => setMenuOpen(false)} />
```
alsuren, line 107:
```suggestion
        <StyledCloseIconWrapper onClick={() => setMenuOpen(false)}>
          <StyledCloseIcon />
```
if you tab to the button and hit space then it doesn't trigger the close action.

It might also be worth renaming StyledCloseIconWrapper to StyledCloseButton, and moving the 
```
  height: 40px;
  width: 40px;
```
inside the styling for the button. That would make this bit look like:
```suggestion
        <StyledCloseButton onClick={() => setMenuOpen(false)} >
          <CloseIcon />
```

frontend/components/Header/NavHeader.tsx
```diff
@@ -89,30 +90,78 @@ const StyledNav = styled.div`
...
-    :hover a {
+    a {
+      color: ${theme.colorNhsukBlue};
+    }
+    :hover a, :hover {
       color: ${theme.colorNhsukWhite};
+      background-color: ${theme.colorShadeNhsukBlue35};
+    }
+    :active {
+      border-bottom: 4px solid ${theme.colorNhsukBlack};
```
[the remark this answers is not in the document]
declanslevin: I logged the tabbing issue for the workspaces directory as a bug: https://github.com/FutureNHS/futurenhs-platform/issues/286

frontend/components/Header/NavHeader.tsx
```diff
@@ -0,0 +1,153 @@
...
+      display: flex;
+
+      li {
+        display: flex;
+      }
+    }
+  `}
+`;
+
+const navItems = [
```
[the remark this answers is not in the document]
alsuren: actually it's better where it is.

frontend/components/Header/Nav/NavList.tsx
```diff
@@ -0,0 +1,100 @@
...
+`;
+
+type NavItems = { title: string; icon: ReactNode; href: string };
+
+interface NavListProps {
+  navItems: NavItems[];
+  setMenuOpen: (open: boolean) => void;
+}
+
+const NavList = ({ navItems, setMenuOpen }: NavListProps) => {
```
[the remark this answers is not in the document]
alsuren: once it is nested inside NavHeader it actually feels fine. Leave it as it is.

frontend/components/Header/NavHeader.tsx
```diff
@@ -0,0 +1,153 @@
...
+  {
+    title: "My workspaces",
+    icon: <WorkspacesIcon />,
+    href: "/workspaces/directory",
```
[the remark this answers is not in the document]
alsuren: ... or just add a comment here that mirrors the one in NavList.tsx?

frontend/components/Icon/DashboardIcon.tsx
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+
+interface Props {
+  className?: string;
+}
+
+const DashboardIcon = ({ className }: Props) => (
+  <div className={`icon-wrapper dashboard-icon-wrapper ${className || ""}`}>
```
frigus02, line 8: Do we need the `dashboard-icon-wrapper` class? It doesn't seem to be used anywhere. Same with the other special classes in the other icons.

frontend/components/NavHeader/NavHeader.tsx
```diff
@@ -0,0 +1,183 @@
+import React, { useState } from "react";
+
+import { Header } from "nhsuk-react-components";
```
frigus02, line 3: What are we getting from using the NHS UK `Header` component? It looks like it's basically just a `<header />` element with a bit of styling.

The reason I'm asking is:
- We're not using it as intended. So we have a small risk of our header breaking when the component updates.
- We're importing quite a bit of CSS for it and we're not using much of it.

frontend/components/NavHeader/NavHeader.tsx
```diff
@@ -0,0 +1,183 @@
...
+      <StyledNavContainer>
+        <NavMenuListItem
+          className="nav-bar-item"
+          title="My workspaces"
+          icon={<WorkspacesIcon />}
+          href="/workspaces/directory"
+        />
+        <NavMenuButton menuOpen={menuOpen} setMenuOpen={setMenuOpen} />
+      </StyledNavContainer>
+      {menuOpen && <NavList navItems={navItems} setMenuOpen={setMenuOpen} />}
```
frigus02, line 178: In theory we could make this menu work without JavaScript by using either the [`<details>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) element or an invisible checkbox.

But this an improvement we should do later when we need to.

frontend/components/NavHeader/NavList.tsx
```diff
@@ -0,0 +1,122 @@
...
+  setMenuOpen: (open: boolean) => void;
+}
+
+const NavList = ({ navItems, setMenuOpen }: NavListProps) => {
+  return (
+    <StyledNavMenuContainer>
+      <p>
+        <span>Menu</span>
+        <StyledCloseIconWrapper onClick={() => setMenuOpen(false)}>
+          <CloseIcon />
```
frigus02, line 103: We should add some alternative text here. It doesn't look like the `CloseIcon` has a prop for that. So let's add an `aria-label` to the `StyledCloseIconWrapper`, e.g. `ariaLabel="close menu"`.

frontend/components/NavHeader/NavList.tsx
```diff
@@ -0,0 +1,122 @@
...
+        </StyledCloseIconWrapper>
+      </p>
+      <ul>
+        {navItems.map((item) => {
+          return (
+            <NavMenuListItem
+              title={item.title}
+              icon={item.icon}
+              href={item.href}
+              key={uuid()}
```
frigus02, line 113: A UUID is not a great key, because it's a random value. The key is used to check if DOM nodes can be reused/reordered when the content changes. It's good to keep them stable. I think the `title` should actually be unique here, so let's use that.
```suggestion
              key={item.title}
```

frontend/components/NavHeader/NavList.tsx
```diff
@@ -0,0 +1,122 @@
...
+      background-color: ${theme.colorNhsukYellow};
+      border-bottom: 4px solid ${theme.colorNhsukBlack};
+      svg {
+        fill: ${theme.colorNhsukBlack};
+      }
+    }
+  `}
+`;
+
+const StyledNavMenuContainer = styled.div`
```
frigus02, line 41: We could make this a `nav` instead of a `div` to give it some semantic meaning. I'm nitpicking, though. It's not too important.

frigus02 (commented):
Looks really good. Most of my suggestions should be small. If anything of this would result in a lot of work, let's ignore it 🙂

frontend/components/Icon/DashboardIcon.tsx
declanslevin: I added in a specific class for each icon in addition to the `icon-wrapper` class in case it becomes necessary to target a specific icon for styling that isn't universal to all of the icons

frigus02: Makes sense. I'm generally in favour of not future proofing this way, unless we know for sure that we will need it exactly like this. In my experience more often then not, we end up removing/refactoring it later because it doesn't quite fit. 🙂

It's a super small thing here, though. I'd say just leave it.

frontend/components/NavHeader/NavHeader.tsx
declanslevin: It might not be necessary to use the NHS uk Header anymore, it's kind of a hangover from the initial build of this component because I was using a lot more of the Nhs Header components to begin with. Besides the Header container the only other thing I'm using is the Nhs logo

frigus02: Yup, makes sense. If it's easy, I would be in favour of removing the dependency on the NHS UK `Header`. If it turns out to be another rabbit hole because we rely on some of the styles, leave it like this.

frontend/pages/_app.scss
```diff
@@ -9,3 +9,4 @@
 @import "node_modules/nhsuk-frontend/packages/components/label/_label";
 @import "node_modules/nhsuk-frontend/packages/components/panel/_panel";
 @import "node_modules/nhsuk-frontend/packages/components/textarea/_textarea";
+@import "node_modules/nhsuk-frontend/packages/components/header/_header";
```
frigus02, line 12: I think that's not needed anymore.

frontend/components/NavHeader/NavHeader.tsx
```diff
@@ -0,0 +1,201 @@
...
+`;
+
+// const StyledHeaderLogo = styled(Header.Logo)`
+//   display: none;
+//   ${({ theme }) => `
+//     @media (min-width: ${theme.mqBreakpoints.largeDesktop}) {
+//       display: block;
+//     }
+//   `}
+// `;
```
frigus02, line 85: We can probably remove this now.

frontend/pages/_app.scss
declanslevin: Turns out this is needed, for the sake of speed I copied classes from the NHS UK logo to get it to style the same

frigus02: Ahhh.

frigus02 (approved):

frigus02:
Looks great. Just merge conflicts now. Hopefully not too hard given that it's mostly snapshots.

PR #257, opened 2020-09-22
Updated with License details

simplybenuk:
Update to README licensing in line with the [GDS-Way](https://github.com/alphagov/gds-way/blob/master/README.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,3 +23,15 @@ The following are used within CI to verify formatting. You may find it useful to
 - [eslint](https://eslint.org/)
 
 Note: Prettier formats all relevent code in this repo, and therefore is installed at the top level. Run `yarn` to install.
+
+## Licence
+
+Unless stated otherwise, the codebase is released under [the MIT License][mit].
+This covers both the codebase and any sample code in the documentation.
+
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -23,3 +23,15 @@ The following are used within CI to verify formatting. You may find it useful to
 - [eslint](https://eslint.org/)
 
 Note: Prettier formats all relevent code in this repo, and therefore is installed at the top level. Run `yarn` to install.
+
+## Licence
+
+Unless stated otherwise, the codebase is released under [the MIT License][mit].
+This covers both the codebase and any sample code in the documentation.
```
alsuren, line 30: @simplybenuk That reminds me that I set the `frontend` folder to be [ISC](https://en.wikipedia.org/wiki/ISC_license) licensed because I was using some example code from nextjs that was ISC licensed in a single file. I will push a commit to tidy things up so that it's only that file that's ISC and the rest is MIT.

alsuren: It turns out https://github.com/vercel/next.js/pull/16691 has made everything MIT, so that makes my life simpler.

frigus02 (approved):

PR #262, opened 2020-09-23
AB1794 - upload single file

### LINKED ISSUE #235: File upload discussion
## File upload discussion

Raising this issue as a place for us to discuss how we upload files to the platform.

This is an initial suggestion of how we might do this:

image

1. The browser makes an authorised HTTPS request to get a Shared Access Signature (SAS), which is, effectively, a URL to which they can upload a temporary file. There are 3 types of SAS. (see https://docs.microsoft.com/en-us/rest/api/storageservices/authorize-requests-to-azure-storage). The documentation recommends that we use a User delegation SAS, but I think this puts a hard dependency on Azure AD (which may be fine for now, but maybe not long term). Next best is a Service SAS, which requires the service to know the Storage Account key. We can specify an upload blob (and container). We can also specify either start time, expiry time and permissions (for an ad-hoc SAS) or link it to a storage access policy on the target storage container (which could allow us to revoke all the tokens if we need to).

2. The browser uploads the file to the URI returned in step 1. We can store metadata against the uploaded file.

3. The browser calls a "publish" endpoint on the files service, which performs some checks on the uploaded file and moves it to a final destination (probably another container, potentially in another storage account). When this is complete it publishes one or more events to signify that the file is now available.

StuartHarris:
## Adds single file upload

Note: merge #266 first!
----
See this issue for more details #235 

- [x] introduce `structopt` to structure args/env config gathering and provide help etc
- [x] add `make run` to fetch the database URL and run workspace service locally
- [x] add `sas` module` to generate a SAS URL for uploading files to temporary container
- [x] integrate with graphql Schema

_Note: the UI for file upload will be in another PR_

[Ticket](https://dev.azure.com/futurenhs/FutureNHS/_workitems/edit/1794/)

## Pre-review checklist:

- [x] Tests

## Testing information

To test in your dev cluster you need to run terraform apply to create the Storage Account and related K8s secret and configmap.

```bash
OVERWRITE=localfiles infrasructure/scripts/create-dev-environment.sh $FNHSNAME
```

- [x] Tested locally

- [x] Tested in cluster


StuartHarris:
I got an email from "GitGuardian" letting me know I've uploaded a private key for Azure Blob Storage. This is true, but I revoked the key before committing it. A real key is needed for the tests, but because it is revoked it can't be used, so I think it's fine. Does anyone have an opinion on whether this is OK?

workspace-service/src/config.rs
```diff
@@ -0,0 +1,53 @@
...
+    /// Key for the Azure EventGrid topic
+    #[structopt(long, env = "EVENTGRID_TOPIC_KEY", hide_env_values = true)]
+    pub eventgrid_topic_key: Option<String>,
+
+    /// The Azure Blob Storage Account key for file uploads
+    #[structopt(
+        long,
+        env = "UPLOAD_MASTER_KEY",
+        hide_env_values = true,
+        required_unless("selfcheck-only")
```
alsuren, line 41: If this is merged into master, will it actually run, or do we need to think about sealed secrets||terraform as part of this PR?

alsuren (commented):
I've not tested locally, but your demo from yesterday looked good, and the code looks good. I guess the last thing to do before merging is check that it's deployable. Trying `./infrastructure/scripts/install-argo-cd.sh dev-$FNHSNAME pr-AB1794-upload-single-file` now.

I suspect that I will end up paying a bunch of money into the acronym swear jar for saying SAS and not remembering to say "shared access signatures".

workspace-service/src/config.rs
StuartHarris: Yeah! Good catch, thanks. Will fix

StuartHarris: Should be good now. I tested it in my cluster. Obvs need to run `terraform apply` first.

alsuren (approved):

PR #263, opened 2020-09-23
develop against local postgresql

alsuren:
something that we needed while pairing.


edcompton (approved):
👏 👏 👏 👏

workspace-service/Makefile
```diff
@@ -1,6 +1,8 @@
 app := workspace-service
 image := fnhsproduction.azurecr.io/$(app)
 tag := latest
+localdb := postgres://postgres:postgres@localhost:5432
+db := $(kubectl get secret -n workspace-service workspace-service -o jsonpath='{.data.url}' | base64 -D)
```
alsuren, line 5:
```suggestion
db := $$(kubectl get secret -n workspace-service workspace-service -o jsonpath='{.data.url}' | base64 -D)
```
My instinct says this but I've not checked it.

workspace-service/Makefile
```diff
@@ -27,19 +29,25 @@ docker-push: ## Push Docker image
 	docker push $(image)
 	docker push $(image):$(tag)
 
+.PHONY: run-local
+run-local: ## Create a local, migrated DB for dev using a local DB rather than cluster, then cargo watch
```
alsuren, line 33:
```suggestion
watch-local: ## Create a local, migrated DB for dev using a local DB rather than cluster, then cargo watch
```

workspace-service/Makefile
```diff
@@ -27,19 +29,25 @@ docker-push: ## Push Docker image
...
+	docker-compose up -d \
+	&& until env DATABASE_URL=$(localdb) cargo sqlx migrate run; do sleep 1; done \
+	&& cargo watch -x help -x check
```
alsuren, line 36: ... there's something strange going on here. If I break one of our .sql files with this running, it doesn't blow up in my face. Something for tomorrow, I think.

workspace-service/Makefile
edcompton: yeah I'd say so

workspace-service/Makefile
```diff
@@ -27,19 +29,25 @@ docker-push: ## Push Docker image
 	docker push $(image)
 	docker push $(image):$(tag)
 
+.PHONY: run-local
```
alsuren, line 32:
```suggestion
.PHONY: watch-local
```
oops. Managed to get them out of sync with my previous suggestion.

workspace-service/Makefile
alsuren: ... ah. the `env DATABASE_URL=$(localdb)` only applies to the `cargo sqlx migrate run` and not the `cargo watch`. That might be the problem.

edcompton: should be sorted now.

PR #265, opened 2020-09-23
PoC inlne-react-svg babel transform

alsuren:
This is a proof of concept of getting babel to load our svgs for us.

There is a bunch of ugliness documented on  https://github.com/twopluszero/next-images/issues/15 but there seem to be workarounds available.

It would mean that the workflow for importing from figma is:

1) export as svg
2) import into git and change the hard coded color to currentColor
3) use `import` to magically import it as a react component

It may be that (2) can be fixed by tweaking the babel transform or something. I'm not sure.

require() still works as before. I guess babel doesn't touch require calls?

I'm also not sure what happens if you require an svg with currentColor as one of the path colors.


alsuren:
I think I needed to do that spike to realise what the downsides of that approach are. After sleeping on it, I agree with you.
Upsides: you just import an svg and get an inline svg react component
Downsides:
1) Caving import behave differently from require() for the same file type is a bit insane. Currently import is forbidden for svgs so it's quite simple. Adding hacks like this make things difficult to reason about
2) As always, we don't just have one javascript buildsystem, we have 3-4:
* nexjs = webpack + babel
* jest = babel + some custom rules
* vscode = tsc

I guess this just doesn't pass the "does this make our lives simpler?" test, so hand-inlining the svgs into one react component per svg gets my vote.

PR #267, opened 2020-09-24
Create folder

trcywu:
[Ticket](https://dev.azure.com/futurenhs/FutureNHS/_boards/board/t/FutureNHS%20Team/Stories/?workitem=1788)

[Figma](https://www.figma.com/file/jlVLY0iKGJdCHPoOJdTIyH/Folders?node-id=128%3A835)

## Acceptance Criteria
image

## Pre-review checklist:

- [ ] Tests

- [ ] Documentation

- [ ] Feature flag

## Testing information

- Test plan?

## Test:

- [ ] Tested locally

- Platforms/Browsers:

  - [ ] Internet Explorer 11

  - [ ] Edge

- Accessibility:

  - [ ] Screen Reader

  - [ ] Keyboard Navigation

  - [ ] Magnification

  - [ ] Contrast

  - [ ] Text size 200%

  - [ ] Landscape (Mobile)


frontend/components/Navigation/Navigation.tsx
```diff
@@ -55,7 +61,25 @@ interface Props {
...
+    description: "create folder",
+    workspace: workspace.id,
+  };
+
+  const icons: { [key: string]: string } = {
+    closed: require("../../public/folderClosed.svg"),
+    open: require("../../public/folderOpen.svg"),
+  };
+
+  const alphabetisedFolders = folders.sort((a, b) => {
```
alsuren, line 77: [The sort() method sorts the elements of an array *in place* and returns the sorted array.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/sort)

are we sure it's safe to mutate the `folders` prop?

frontend/components/NavListItem/NavListItem.tsx
```diff
@@ -49,16 +51,17 @@ const ListItem = styled.li<{ active: boolean }>`
...
+  item,
+  imgSrc,
+  className,
+  href,
+}: Props) => (
+  <ListItem active={active} className={className}>
+    <Link href={href}>
       <a>
-        <img src={active ? icons["open"] : icons["closed"]} />
+        <img src={imgSrc} />
```
frigus02, line 64: That `img` should have an `alt` text. I see it didn't have one before, either. If it's not too much work, you could add it now, or we can make another PR for that.

For the folder, I think the icon doesn't add any additional information, so it should probably be `alt=""` (empty string).

For the toggled state (open/closed), the icon does convey additional information. So the text should probably contain that, e.g. `alt="expanded"`/`alt="collapsed"`.

frontend/components/Navigation/Navigation.tsx
```diff
@@ -1,8 +1,7 @@
-import React from "react";
+import React, { FC } from "react";
 
 import Link from "next/link";
 import styled from "styled-components";
-import { v4 as uuid } from "uuid";
```
frigus02, line 5: I think this was the last usage of `uuid`. So we can remove it from the package.json.

frontend/components/Navigation/Navigation.tsx
```diff
@@ -71,12 +95,22 @@ const Navigation = ({ workspace, folders, activeFolder }: Props) => {
...
+            imgSrc={require("../../public/createFolder.svg")}
+            className="nav-list-item"
+            href={`/workspaces/${workspace.id}/folders/${createFolder.id}`}
+          />
+          {alphabetisedFolders.map((folder) => (
             <NavListItem
-              active={folder.id == activeFolder}
-              key={uuid()}
+              active={folder.id === activeFolder}
+              key={folder.id}
```
frigus02, line 108: It's nice to use a stable ID for the `key`. ❤️

frontend/components/Navigation/Navigation.tsx
frigus02: Hm, good question. I guess the alternative is to create a new array and sort it, e.g. `[...folders].sort((a, b) => {})`

frontend/components/Navigation/Navigation.tsx
```diff
@@ -55,7 +61,25 @@ interface Props {
...
+
+  const icons: { [key: string]: string } = {
+    closed: require("../../public/folderClosed.svg"),
+    open: require("../../public/folderOpen.svg"),
+  };
+
+  const alphabetisedFolders = folders.sort((a, b) => {
+    const textA = a.title.toUpperCase();
+    const textB = b.title.toUpperCase();
+    return textA < textB ? -1 : textA > textB ? 1 : 0;
```
frigus02, line 80: These are strings, so you could use [localeCompare](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/localeCompare) instead:
```suggestion
    return textA.localeCompare(textB);
```

Also I just learned that this function supports case-insensitive comparisons. So you can replace this entire sorting logic with:

```
folders.sort((a, b) => a.localeCompare(b, 'en', { sensitivity: 'base' }))
```

frontend/pages/workspaces/[workspaceId]/folders/create-folder.tsx
```diff
@@ -0,0 +1,194 @@
...
+import { requireAuthentication } from "../../../../lib/auth";
+import { getSdk, Folder } from "../../../../lib/generated/graphql";
+
+export const getServerSideProps: GetServerSideProps<Props> = requireAuthentication(
+  async (context) => {
+    const client = new GraphQLClient(
+      "http://workspace-service.workspace-service/graphql"
+    );
+    const sdk = getSdk(client);
+    const workspaceId = (context.params?.workspaceId as string) || "";
```
Comment (frigus02) [inline frontend/pages/workspaces/[workspaceId]/folders/create-folder.tsx:24]: I feel like we should throw an error if the workspaceId is not set. It will error anyway because we won't be able to find a workspace. I think throwing an error explicitly makes this clearer.

frontend/pages/workspaces/[workspaceId]/folders/create-folder.tsx
```diff
@@ -0,0 +1,194 @@
...
+      const client = new GraphQLClient("/api/graphql");
+      const sdk = getSdk(client);
+      const newFolder = await sdk.CreateFolderMutation(newFolderDetails);
+      const {
+        createFolder: { id: folderId },
+      } = newFolder;
+      router.push(`/workspaces/${workspaceId}/folders/${folderId}`);
+    } catch (error) {
+      console.log("Create folder failed", error);
+      window.alert("Error creating folder, failed");
```
Comment (frigus02) [inline frontend/pages/workspaces/[workspaceId]/folders/create-folder.tsx:108]: Hm, I'm not sure what we should do in the error case. On the workspace creation page I think an alert is okay, because it's only visible to the service team. But this is for the end user. A proper error message would be nicer. Could we display an error message on the page instead? It looks like the NHS UK React components have an `ErrorMessage` component. I assume the default styling should be fine.

frigus02 (commented):
I realized I left quite a few comments. But it looks really good actually 🙂

frontend/components/NavListItem/NavListItem.tsx
trcywu: Yes my bad, should have spotted this. I think for the folder we can still have 'create new folder icon' as it should describe what the image is instead of leaving it blank. For the open and close I think 'open folder icon' and 'close folder icon' instead of expanded or collapsed? Otherwise it signals accordion style as if there's something under the NavListItem that you can read which we don't. The 2 folder states only signal whether we are on the folder or not.

frontend/components/Navigation/Navigation.tsx
trcywu: i can do the [...folders] as Jan suggested below. What's the risk of mutating folders straight up?

frontend/components/NavListItem/NavListItem.tsx
frigus02: No worries.

I don't think specifying "create new folder icon" helps. It doesn't add any information because the text "New folder" is right next to it. It would just duplicate the information, which doesn't help.

You can also think about it this way: would you still understand what the menu item is about if we removed the icon? I would say yes. This means the icon is only there to make it visually nicer. And in this case it's correct to set `alt=""`. It tells screen readers to ignore the image.

For open and close: You're right. I got the meaning wrong. It's showing us that this is the active menu item. In this case, that is what the alt text should be. I think I would set it to:
- closed: `alt="folder"` (We could do something like: "Type: folder". I think the purpose here is to indicate the type of content, e.g. folder, file, ...).
- open: `alt="folder (current page)"` (I don't think it matters that the icon shows an open folder. The important thing is that it indicates that we are currently on this folder.)

frontend/components/Navigation/Navigation.tsx
frigus02: It's not a problem at the moment, but could be unexpected. Imagine having a component like this (ignore that it's super pointless 😄):

```tsx
const MyPage = () => {
  const folders = [
    "B", "A"
  ];
  return (
    <div>
      <Navigation folders={folders} />
      <div>first: {folder[0]}</div>
    </div>
  )
};
```

You probably wouldn't expect the navigation to sort the folders and the next div to print "A".

It's unlikely to actually be a problem for us. But you can change it to `[...folders]`, just to be safe. When someone looks at it later on, it's clear that the ordering is purely internal in the component.

frontend/components/Navigation/Navigation.tsx
frigus02: Well, I just realized I forgot the `title` property. It should be:
```ts
folders.sort((a, b) => a.title.localeCompare(b.title, 'en', { sensitivity: 'base' }))
```

frontend/components/Navigation/Navigation.tsx
trcywu: ok i understand.

frontend/components/NavListItem/NavListItem.tsx
trcywu: ok cool this makes sense

frontend/pages/workspaces/[workspaceId]/folders/create-folder.tsx
Comment (trcywu) [reply frontend/pages/workspaces/[workspaceId]/folders/create-folder.tsx:108]: Should i have the ErrorMessage displayed under the Submit & discard buttons?

Comment (frigus02) [reply frontend/pages/workspaces/[workspaceId]/folders/create-folder.tsx:108]: Good question 😅. Yeah, that sounds fine to me. Can always improve later on.

frontend/pages/workspaces/[workspaceId]/folders/create-folder.tsx
```diff
@@ -0,0 +1,194 @@
...
+interface Props {
+  workspaceFolders: Array<Pick<Folder, "id" | "title">>;
+  workspace: Pick<Workspace, "id" | "title">;
+}
+
+const CreateFolder: FC<Props> = ({ workspaceFolders, workspace }) => {
+  const [remainingChars, setRemainingChars] = useState({
+    title: null,
+    description: null,
+  });
```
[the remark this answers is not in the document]
Comment (trcywu) [reply frontend/pages/workspaces/[workspaceId]/folders/create-folder.tsx:89]: I might re-examine this another time and not in this PR as i predict some refactoring in the future.

frontend/components/Navigation/Navigation.tsx
```diff
@@ -71,12 +93,26 @@ const Navigation = ({ workspace, folders, activeFolder }: Props) => {
...
       <NavSection title="Folders">
         <List>
-          {folders.map((folder) => (
+          <NavListItem
+            active={createFolder.id === activeFolder}
+            item={createFolder}
+            imgSrc={require("../../public/createFolder.svg")}
+            className="nav-list-item"
+            href={`/workspaces/${workspace.id}/folders/${createFolder.id}`}
+            altText="create new folder icon"
```
frigus02, line 102: I'm quite confident that it would be better with an empty string as the alt text here, in order to not duplicate information. But I'm also not 100% sure, so happy to leave like this, if you think it's better. 🙂

infrastructure/environments/dev/backend-config.tfvars
```diff
@@ -0,0 +1,2 @@
+resource_group_name="tfstatetwu"
+storage_account_name="fnhstfstatedevtwu"
```
frigus02, line 2: That should probably not be here. I noticed we forgot to add this file to the gitignore.

frontend/pages/workspaces/[workspaceId]/folders/create-folder.tsx
Comment (frigus02) [reply frontend/pages/workspaces/[workspaceId]/folders/create-folder.tsx]: Looks great now 🎉

frontend/components/Navigation/Navigation.tsx
trcywu: i'm good with empty string! Thought i had changed it already. Will do now

infrastructure/environments/dev/backend-config.tfvars
trcywu: ill add to gitignore

frigus02: Thank you ❤️

frigus02 (approved):
🎉

alsuren (approved):

frontend/components/Navigation/Navigation.tsx
frigus02: Oops, I just saw that we're not doing `[...folders].sort`, yet. Possibly because I distracted you with the `localeCompare` thing 🙈. Do you want to do that as a last fix?

PR #269, opened 2020-09-25
Run Cypress in CI with mocks.

edcompton:
[Ticket](https://dev.azure.com/futurenhs/75407963-f812-43e5-a734-1059bbc036a3/_workitems/edit/2176)

## Acceptance Criteria

- cypress tests for chrome and firefox are run in CI
- Login is mocked so no need to call Azure AD
- GraphQL calls are mocked via a stub server

Screenshot 2020-09-30 at 16 43 47

Screenshot 2020-09-30 at 14 18 15

## Pre-review checklist:

- [x] Tests

- [x] Documentation

- [ ] ~~Analytics (user analytics, e.g. Google Analytics)~~

- [ ] ~~Observability (metrics/tracing)~~

- [ ] ~~Feature flag~~

## Testing information

environment variables need to be added to `.env` files;

`WORKSPACE_SERVER_API_ROOT=http://localhost:3001/graphql` should be added to `.env.development.local` if you want to use the responses listed in the `cypress/fixtures` directory, rather than port-forwarding the workspace service when working locally. Else just use `http://workspace-service.workspace-service/graphql` as a value. 

- Test plan?

## Test:

- [x] Tested locally

- [x] Tested in CI 

- Platforms/Browsers:

  - [ ] ~~Internet Explorer 11~~

  - [ ] ~~Edge~~


frontend/lib/auth.ts
```diff
@@ -20,7 +20,16 @@ export const requireAuthentication = <P>(
...
+  const dev = process.env.NODE_ENV !== "production";
+
+  if (dev) {
+    const user = {
+      id: "1234",
+      name: "Test user",
+      emails: ["test@example.com"],
+    };
+    return getServerSideProps(context, user);
+  } else if (!user) {
```
frigus02, line 32: We should be able to get around that by adding `redirectAuthSuccess` to the route handler of `/auth/login` in server.js.

frontend/package.json
```diff
@@ -6,6 +6,7 @@
   "scripts": {
     "predev": "yarn generate",
     "dev": "NODE_ENV=development node server.js",
+    "dev:local": "yarn start:stubs & NODE_ENV=development node server.js",
```
frigus02, line 9: I would say let's make this the default `dev` command.

However. We need to ensure the stubs server stops properly. We could use https://www.npmjs.com/package/concurrently for example.

frontend/pages/workspaces/[workspaceId].tsx
```diff
@@ -14,9 +14,9 @@ import { getSdk, Folder } from "../../lib/generated/graphql";
 
 export const getServerSideProps: GetServerSideProps<Props> = requireAuthentication(
   async (context) => {
-    const client = new GraphQLClient(
-      "http://workspace-service.workspace-service/graphql"
-    );
+    const workspaceUrl: string =
+      process.env.WORKSPACE_SERVER_API_ROOT || "http://localhost:3000/graphql";
```
Comment (frigus02) [inline frontend/pages/workspaces/[workspaceId].tsx:18]: Maybe we want a little helper function like:
```ts
const requireEnv = (name: string): string => {
  const value = process.env[name];
  if (!value) {
    throw new Error(`Environment variable ${name} is required`);
  }

  return value;
};
```

frontend/README.md
```diff
@@ -26,3 +26,15 @@ yarn dev
 Open [http://localhost:3000](http://localhost:3000) with your browser to see the result.
 
 You can start editing the page by modifying `pages/index.js`. The page auto-updates as you edit the file.
+
+## Testing
+
+### Cypress
+
+Cypress tests for chrome and firefox currently run as part of our pipeline. It's also possible to run in a local browser before pushing code.
```
frigus02, line 34:
```suggestion
Cypress tests run as part of our CI pipeline. It's also possible to run in a local browser before pushing code.
```

frontend/.env.development
```diff
@@ -1,3 +1,2 @@
 ORIGIN=http://localhost:3000
 COOKIE_SECRET=below-particular-sky-mix
-AAD_B2C_CLIENT_ID=3d007909-ddef-4c9d-9e2a-cf4e6b4b8753
```
frigus02, line 3: Should we instead put `WORKSPACE_SERVER_API_ROOT=http://localhost:3001` in here? I think this might then also remove the need to specify this in the GitHub Action workflow.

frontend/manifests/production/config-map.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
+kind: ConfigMap
```
frigus02, line 2: I think we should move this to the `base` folder. Dev clusters behave the same as prodution, so they need the same value for the environment variable.

frontend/lib/auth.ts
edcompton: 👍

frontend/package.json
edcompton: done on both counts

frontend/pages/workspaces/[workspaceId].tsx
Comment (edcompton) [reply frontend/pages/workspaces/[workspaceId].tsx]: 👍

frontend/.env.development
edcompton: Good idea!

frontend/manifests/production/config-map.yaml
edcompton: 👍

frigus02 (commented):
Looks great. I think there is one remaining bit of confusion and I also blame myself for the sloppy comment before.

The `WORKSPACE_SERVER_API_ROOT` env var should either be just the origin or the full URL. We have a mix at the moment. Seems good to me to use the full URL. So that means:

- Change .env.development to use http://localhost:3001/graphql
- Change frontend/lib/withUrqlClient.ts to not append "/graphql" but use the env var as is

frigus02 (approved):

frontend/manifests/base/config-map.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: config_map
+data:
+  workspace_server_api_root: http://workspace-service.workspace-service/graphql
```
alsuren, line 6: workspace service graphql endpoint?

frontend/package.json
```diff
@@ -14,7 +14,10 @@
     "test": "jest",
     "lint": "yarn eslint . --ext .js,.ts,.tsx",
     "clean": "rm -rf .next",
-    "generate": "graphql-codegen --config codegen.yaml"
+    "generate": "graphql-codegen --config codegen.yaml",
+    "cypress:local": "concurrently \"yarn start:stubs\" \"NODE_ENV=development node server.js\" \"cypress open --config-file ./cypress.json\"",
+    "start:stubs": "node stubServer",
+    "cypress:ci": "yarn start:stubs & yarn dev"
```
alsuren, line 20:
```suggestion
    "start:stubs": "node stubServer"
```
and just use yarn dev in ci

frontend/package.json
```diff
@@ -76,11 +79,16 @@
...
-    "eslint": "^7.10.0",
-    "eslint-plugin-import": "^2.22.1",
-    "eslint-plugin-react": "^7.21.2",
+    "concurrently": "^5.3.0",
+    "cypress": "^5.2.0",
+    "eslint": "^7.9.0",
+    "eslint-plugin-cypress": "^2.11.1",
+    "eslint-plugin-import": "^2.22.0",
+    "eslint-plugin-react": "^7.20.6",
+    "express-graphql": "^0.11.0",
```
alsuren, line 91: kill this

frontend/stubServer/index.js
```diff
@@ -0,0 +1,38 @@
...
+    if (err) {
+      throw err;
+    }
+    console.log(`> Ready on http://localhost:${port}`);
+  });
+};
+
+(() => {
+  stubServer();
+})();
```
alsuren, line 38: just call it directly

frontend/stubServer/index.js
```diff
@@ -0,0 +1,38 @@
...
+  path.join(__dirname, "schema", "../../schema.graphql"),
+  "utf8"
+);
+const typeDefs = gql`
+  ${schemaCode}
+`;
+
+const stubServer = () => {
+  app.use(bodyParser.json({ limit: "10mb" }));
+  app.use(bodyParser.urlencoded({ extended: true }));
```
alsuren, line 21: can AppolloServer do without these?

frontend/stubServer/index.js
```diff
@@ -0,0 +1,38 @@
...
+const resolvers = require("./resolver");
+
+const port = 3001;
+const schemaCode = fs.readFileSync(
+  path.join(__dirname, "schema", "../../schema.graphql"),
+  "utf8"
+);
+const typeDefs = gql`
+  ${schemaCode}
+`;
```
alsuren, line 17: const typeDefs = gql(schemaCode)

frontend/stubServer/resolver.js
```diff
@@ -0,0 +1,31 @@
+const fs = require("fs");
+const path = require("path");
+
+const folderByWorkspaceResponse = require("../cypress/fixtures/folder-by-workspace-graphql-response.json");
+const folderResponse = require("../cypress/fixtures/folder-graphql-response.json");
+const workspaceResponse = require("../cypress/fixtures/workspace-graphql-response.json");
+const workspacesResolver = {
```
alsuren, line 7: newline

alsuren (approved):

PR #270, opened 2020-09-25
Folder created event

alsuren:
[AB#2137](https://dev.azure.com/futurenhs/75407963-f812-43e5-a734-1059bbc036a3/_workitems/edit/2137)

This adds the FolderCreated event and emits it during folder creation.

It also adds a mock db::Folder implementation similar to db::Workspace, and does some minor refactors, each in their own commits.

I used https://github.com/FutureNHS/futurenhs-platform/pull/229 as a crib when writing this PR, so the shape should be basically the same.


StuartHarris (approved):
Looks great. Super minor point about `test_helpers.rs` – maybe `test_mocks.rs` or something. Just wary that anything called helpers or utils just ends up being a dumping ground.
Also we should discuss if we need TypeScript support for events, if we only emit from APIs rather than the webserver?

alsuren:
> Also we should discuss if we need TypeScript support for events, if we only emit from APIs rather than the webserver?

Let's nab Jan to chat about this next week. I can't remember what the use-case for typescript was (push notifications maybe?).

PR #271, opened 2020-09-26
Urql

StuartHarris:
This ticket updates the UI to use `urql` instead of `graphql-request` for making GraphQL requests to the server.

It is intended to promote discussion as to whether this is a move the team wants to make. IMO it is better for the following reasons:

- components can be responsible for their own data fetching requirements, so we don't have to pass everything down from the page, through the component hierarchy as props
- queries are aggregated client-side and server-side before being sent to the API
- data is cached so that the same call is not made more than once
   - e.g. server-side 2 components want the same data
   - e.g. client-side the user moves back to a folder they were already in
- if a component's data fetch fails, the error can be local to the component
- we can easily display a component spinner while loading
- code IMO is more idiomatic and simpler to understand

The UI functions correctly as far as I can tell. Please run it locally and have a good look around.

If we decide to use this, I will fix the tests and we should be good to go. :-)


matt-paul:
This looks really nice.

matt-paul:
Out of interest, what puts Urql ahead against other clients with caching abilities?

StuartHarris:
> Out of interest, what puts Urql ahead against other clients with caching abilities?

I think they're pretty comparable and all seem to have improved a lot recently. Not sure if Relay still puts demands on GraphQL format. I think they improved Apollo Client a lot in 3.0, so Urql's lead may be smaller now (it was a lot smaller/faster etc). I think Urlq's extensibility using exchanges is cool. 

Anyway, not married to Urql. But it seems to work well. Especially with SSR. And it's easy to use.

frigus02:
With this we can use hooks and still use server-side rendering? This looks pretty nice.

I don't know if urql of Apollo are better, but I assume it doesn't really make a difference for us. The query code is generated anyway. So I'd say sure. Let's go with this. 👍

frontend/pages/admin/create-workspace.tsx
```diff
@@ -152,4 +141,4 @@ const CreateWorkspace = () => {
   );
 };
 
-export default CreateWorkspace;
+export default withUrqlClient(CreateWorkspace);
```
alsuren, line 144: I'm trying to spot how `useCreateWorkspaceMutation()` invalidates the cache for `useGetWorkspacesQuery()`. It is not obvious.

alsuren: ah. https://formidable.com/open-source/urql/docs/basics/document-caching/ - any mutation affecting Workspace objects will invalidate the cache for all queries that returned Workspace types. That sounds reasonable. The gotcha about empty arrays is a bit scary though.

I think that it will be easier to evaluate this once it has been rebased on top of Tracy's create folder PR.

StuartHarris: See https://formidable.com/open-source/urql/docs/basics/document-caching/. Although I think we can probably do better by using Graphcache. Will investigate this at some point.

alsuren: if only they had a policy that uses cache-first, but falls back to cache-and-network if it spots an empty array or something, to work around the bug.

That said, how bad would it be to use the cache-and-network policy everywhere?

frontend/components/Navigation/Navigation.tsx
```diff
@@ -50,33 +50,39 @@ const List = styled.ul`
...
 }
 
-const Navigation = ({ workspace, folders, activeFolder }: Props) => {
+const Navigation = ({ workspaceId, workspaceTitle, activeFolder }: Props) => {
+  const [{ data, fetching, error }] = useFoldersByWorkspaceQuery({
+    variables: { workspace: workspaceId },
+  });
+
+  if (fetching) return <p>Loading...</p>;
+  if (error) return <p> Oh no... {error?.message} </p>;
```
frigus02, line 64: These should probably use the `ErrorMessage` and some kind of loading component. But we can do that later. It's already better then before.

frontend/pages/workspaces/[workspaceId].tsx
```diff
@@ -54,31 +32,36 @@ const H2 = styled.h2`
...
-      </ContentWrapper>
-    </PageLayout>
-  </>
-);
 
-export default WorkspaceHomepage;
+const WorkspaceHomepage: NextPage = () => {
+  const router = useRouter();
+  const { workspaceId } = router.query;
+  const id = (workspaceId || "unknown").toString();
```
Comment (frigus02) [inline frontend/pages/workspaces/[workspaceId].tsx:42]: I think we should probably throw an error if there is no workspace id in the query. Ideally that would return a `404`. Not sure if that's possible.

frontend/pages/workspaces/[workspaceId]/folders/[folderId].tsx
```diff
@@ -52,30 +30,45 @@ const ContentWrapper = styled.div`
...
-interface Props {
-  folder: Folder;
-  workspaceFolders: Array<Pick<Folder, "id" | "title">>;
-  workspace: Pick<Workspace, "id" | "title">;
-}
+const FolderHomepage: NextPage = () => {
+  const router = useRouter();
+  let { workspaceId, folderId } = router.query;
+  workspaceId = (workspaceId || "unknown").toString();
+  folderId = (folderId || "unknown").toString();
```
Comment (frigus02) [inline frontend/pages/workspaces/[workspaceId]/folders/[folderId].tsx:37]: Similarly, I think these should throw some kind of error.

frontend/__tests__/pages/workspaces/[workspaceId]/folders/[folderId].test.tsx
```diff
@@ -1,31 +1,70 @@
...
-        id: "f1",
-        title: "folder 1",
-        description: "first folder",
-        workspace: "1",
-      },
-    ];
+  const responseState = {
+    executeQuery: jest.fn(({ query }) => {
+      if (
+        query.definitions.find((d: any) => d.name.value == "GetWorkspaceByID")
```
Comment (frigus02) [inline frontend/__tests__/pages/workspaces/[workspaceId]/folders/[folderId].test.tsx:14]: This feels like a common pattern. Might be nice to have a test helper to mock those GraphQL queries. Something like this could be easier to read:

```ts
const responseState = mockGraphQLClient({
  GetWorkspaceByID: {
    workspace: {
      id: "1",
      title: "hospital",
    },
  },
  GetFolderById: {
    folder: {
      id: "f1",
      title: "folder 1",
      description: "first folder",
      workspace: "1",
    },
  },
});
```

But probably something for later as well. Not really related to this PR.

frigus02 (approved):
Works and looks very clean 🎉 . Left a few comments but we don't have to address them here.

frontend/components/Navigation/Navigation.tsx
StuartHarris: Yeah, I agree. Let's do the loading spinners and errors as another PR.

frontend/__tests__/pages/workspaces/[workspaceId]/folders/[folderId].test.tsx
Comment (StuartHarris) [reply frontend/__tests__/pages/workspaces/[workspaceId]/folders/[folderId].test.tsx:14]: Yeah, that would be a great thing to do. It's a shame we have to do any mocking, but now we don't pass the data in as props inevitably means more mocking so we should make it as straightforward as possible.

frigus02 (approved):

PR #280, opened 2020-09-28
Fix frontend event types

frigus02:
The frontend build fails: https://github.com/FutureNHS/futurenhs-platform/runs/1175766764

```
#12 10.58 ./lib/events.ts:20:48
#12 10.58 Type error: Type '{ id: string; subject: string; eventType: "ContentViewed" | "FolderCreated" | "WorkspaceCreated" | undefined; eventTime: Date; data: { [k: string]: unknown; userId: string; contentId: string; contentType: "Folder" | "File"; workspaceId: string; error?: string | undefined; } | { ...; } | { ...; } | undefined; dataVer...' is not assignable to type 'EventGridEvent'.
#12 10.58   Types of property 'eventType' are incompatible.
#12 10.58     Type 'string | undefined' is not assignable to type 'string'.
#12 10.58       Type 'undefined' is not assignable to type 'string'.
#12 10.58 
#12 10.58   18 |     const client = new EventGridClient(new TopicCredentials(key));
#12 10.58   19 |     const topicHostname = new URL(endpoint).host;
#12 10.58 > 20 |     await client.publishEvents(topicHostname, [fullEvent]);
#12 10.58      |                                                ^
#12 10.58   21 |   } else {
#12 10.58   22 |     console.log(`EVENT: ${JSON.stringify(fullEvent)}`);
#12 10.58   23 |   }
```

This fixes the types for now.

We should probably:
- Make sure we catch this on PR-build time
- Make it easier to create an event object

But we can do both of those later.

### DIFF
--- a/.github/workflows/frontend-branch.yaml
+++ b/.github/workflows/frontend-branch.yaml
@@ -6,6 +6,7 @@ on:
       - master
     paths:
       - frontend/**
+      - event-models/**
       - infrastructure/dev-overlay-variables.json
       - .github/workflows/frontend-branch.yaml
 
--- a/.github/workflows/frontend-master.yaml
+++ b/.github/workflows/frontend-master.yaml
@@ -6,6 +6,7 @@ on:
       - master
     paths:
       - frontend/**
+      - event-models/**
       - infrastructure/dev-overlay-variables.json
       - .github/workflows/frontend-master.yaml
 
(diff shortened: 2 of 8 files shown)


alsuren (approved):
I just pushed a change to .github/workflows to make sure we catch regressions in future.

PR #284, opened 2020-09-29
Fix Cypress running on BrowserStack

frigus02:
I noticed Cypress tests fail on BrowserStack. This was caused by #198, which introduced a Cypress plugin to load environment variables from the .env file. This doesn't work on BrowserStack because the .env file is not uploaded.

The fix here:

- Parses the .env file in `cypress-bs.sh` and passes the environment variables as a command line argument to `browserstack-cypress run`.

- Makes use of a new feature and specifies browserstack authentication using environment variables. This means `browserstack.json` no longer contains secrets and we can commit it as is.

- Updates browser versions to the most recent versions in `browserstack.json`.

- Tells BrowserStack through an environment variable if it's supposed to connect to BrowserStackLocal. This means it's no longer required to run `yarn local` when you just want to run `yarn cypress-bs:prod`.

## Testing information

Test Cypress and BrowserStack related commands:
- `yarn cypress-bs:local`
- `yarn cypress-bs:prod`
- `yarn internet-explorer-bs:local`
- `yarn internet-explorer-bs:prod`

## Test:

- [x] Tested locally

- Platforms/Browsers:

  - [x] Internet Explorer 11

  - [x] Edge


frigus02:
This is possibly pointless. We might decide to not run Cypress tests on BrowserStack for simplicity. Instead we run them automated in CI. See #269. In this case we can probably just close this PR and forget about it.

PR #288, opened 2020-09-30
AB1794 upload single file UI

StuartHarris:
Adds UI to upload a single file.

### Note: This also updates Terraform to 0.13.4 (from 0.13.3), so after merging, everyone will need to ensure they are on latest Terraform. 

[1794](https://dev.azure.com/futurenhs/FutureNHS/_workitems/edit/1794/)

[Figma](https://www.figma.com/file/tYFsqhG61CleWS1tUapAar/Source-of-Truth-UI?node-id=258%3A883)

## Acceptance Criteria

- [x] as you hover over a folder in the nav, the meatball appears to the right, in grey
- [x] in the folder page there is a grey meatball in the top right
- [x] as you hover over a meatball, it turns blue
- [x] a tooltip shows after a short delay
- [x] when you click a meatball it turns yellow
- [x] a menu appears
- [x] select "Upload file to this folder"
- [x] show file upload UI
- [x] upload file...

## Pre-review checklist:

- [ ] Tests

- [ ] Documentation

- [ ] Feature flag

- [ ] ~~Data columns adhere to the [Dublin Core Metatdata schema](https://www.dublincore.org/specifications/dublin-core/dcmi-terms/). Further information is available [here](https://www.gov.uk/government/publications/recommended-open-standards-for-government/using-metadata-to-describe-data-shared-within-government)~~ (we cannot store metadata in Azure Blob Storage if the keys do not conform to C# identifiers, so `dc.title` is not valid. Suggest we make sure we are Dublin Core when exposing this data as an API but internal storage doesn't matter (i.e. `title` is fine))

## Testing information

- Test plan?

## Test:

- [ ] Tested locally

- Platforms/Browsers:

  - [ ] Internet Explorer 11

  - [ ] Edge

- Accessibility:

  - [ ] Screen Reader

  - [ ] Keyboard Navigation

  - [ ] Magnification

  - [ ] Contrast

  - [ ] Text size 200%

  - [ ] Landscape (Mobile)


frontend/components/Icon/DeleteIcon.tsx
```diff
@@ -0,0 +1,26 @@
+import React from "react";
+
+interface Props {
+  className?: string;
+}
+
+const Icon = ({ className }: Props) => (
+  <div className={`icon-wrapper logout-icon-wrapper ${className || ""}`}>
```
frigus02, line 8: The class `logout-icon-wrapper` doesn't really fit here. We don't use that class anywhere. Let's remove it from here and the other icons.

frigus02 (commented):
I started reviewing. But need to stop now. Will continue tomorrow.

frontend/components/Icon/DeleteIcon.tsx
StuartHarris: Good spot! Thanks. Copy paste error!

frontend/components/FolderMenu/FolderMenuButton.tsx
```diff
@@ -0,0 +1,79 @@
...
+const MenuButton: FC<MenuProps> = ({
+  className,
+  startHidden,
+  menuOpen,
+  setMenuOpen,
+  ...rest
+}) => {
+  return (
+    <Tooltip tooltip="Options">
+      <button
```
frigus02, line 23: Looks like this button doesn't have any text. Should we give it an `aria-label="Options"`?

frontend/components/FolderMenu/FolderMenuList.tsx
```diff
@@ -0,0 +1,70 @@
...
+    }
+  }
+`;
+
+interface Props extends ComponentPropsWithoutRef<"div"> {
+  startHidden: boolean;
+}
+
+const FolderMenuList: FC<Props> = ({ children, startHidden, className }) => {
+  const items = children as ReactNodeArray;
```
frigus02, line 53: It might make sense to cast `children` directly to `MenuItem[]`. Or maybe we can set `children: MenuItem[]` in the props definition.

frontend/components/FolderMenu/FolderMenuList.tsx
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    li:last-child a {
+      border-radius: 0 0 4px 4px;
+    }
+  }
+`;
+
+interface Props extends ComponentPropsWithoutRef<"div"> {
+  startHidden: boolean;
```
frigus02, line 49: What does `startHidden` do in this component? It looks like it slightly repositions it. What does that do?

frontend/components/NavListItem/NavListItem.tsx
```diff
@@ -59,15 +78,21 @@ const NavListItem: FC<Props> = ({
...
-      </a>
-    </Link>
-  </ListItem>
-);
+  menu,
+}: Props) => {
+  return (
+    <ListItem className={className}>
+      <LinkWrapper active={active}>
+        <Link href={href}>
```
frigus02, line 86: I know we didn't have that before, but should we use `passHref` here, so the `a` gets the href prop?

frontend/pages/workspaces/[workspaceId]/folders/[folderId]/upload-file.tsx
```diff
@@ -0,0 +1,208 @@
...
+                })}
+                aria-invalid={errors.files ? "true" : "false"}
+                error={errors.files && `Please select a file`}
+              />
+              {errors.files?.type === "required" && (
+                <p role="alert">This is required</p>
+              )}
+            </FormField>
+            <p>
+              All uploaded content must conform to to the platform&apos;s{" "}
```
Comment (frigus02) [inline frontend/pages/workspaces/[workspaceId]/folders/[folderId]/upload-file.tsx:187]:
```suggestion
              All uploaded content must conform to the platform&apos;s{" "}
```

frontend/pages/workspaces/[workspaceId]/folders/[folderId]/upload-file.tsx
```diff
@@ -0,0 +1,208 @@
+import React, { useState } from "react";
+
+import { BlockBlobClient } from "@azure/storage-blob";
+import { ErrorMessage as HookFormErrorMessage } from "@hookform/error-message";
+import { NextPage } from "next";
+import { useRouter } from "next/dist/client/router";
+import { Input, Form, Button, ErrorMessage } from "nhsuk-react-components";
+import { useForm } from "react-hook-form";
```
Comment (frigus02) [inline frontend/pages/workspaces/[workspaceId]/folders/[folderId]/upload-file.tsx:8]: Can you merge master into your branch? I found that react-hook-form needs a special import to work in IE11. I added a ESLint rule to enforce that.

Also `@hookform/error-message` doesn't work at all in IE11. Luckily you can easily replace `HookFormErrorMessage` with `{errors.server && <ErrorMessage>{errors.server}</ErrorMessage>}`

frontend/pages/workspaces/[workspaceId]/folders/[folderId]/upload-file.tsx
```diff
@@ -0,0 +1,208 @@
...
+      .then(({ error, data }) => {
+        const { title, files } = formData;
+        if (error) {
+          setError("form", {
+            type: "server",
+            message: error.toString(),
+          });
+        }
+        if (data) {
+          const blobClient = new BlockBlobClient(data.fileUploadUrl);
```
Comment (frigus02) [inline frontend/pages/workspaces/[workspaceId]/folders/[folderId]/upload-file.tsx:104]: I wonder if we have to use `@azure/storage-blob` package or if we could shave off a bit of JavaScript and do a form `POST` to the blob URL to upload the file. But for the benefit of speed, we should probably just accept that and move on.

frontend/pages/workspaces/[workspaceId]/folders/[folderId]/upload-file.tsx
```diff
@@ -0,0 +1,208 @@
...
+            .uploadBrowserData(files[0], {
+              maxSingleShotSize: 4 * 1024 * 1024,
+            })
+            .then(() => {
+              blobClient
+                .setMetadata({ title, filename: files[0].name })
+                .then(() => {
+                  router.push(`/workspaces/${workspaceId}/folders/${folderId}`);
+                });
+            });
```
Comment (frigus02) [inline frontend/pages/workspaces/[workspaceId]/folders/[folderId]/upload-file.tsx:115]: Do we want some error handling in case this fails?

frigus02 (commented):
Looks really good to me. I feel like most of my comments are nitpicking on the `startHidden` prop. 🙂

frontend/components/FolderMenu/FolderMenu.tsx
```diff
@@ -0,0 +1,124 @@
...
+  LockIcon,
+  MeatballIcon,
+  MoveIcon,
+  UploadIcon,
+} from "../Icon";
+import FolderMenuList from "./FolderMenuList";
+import { MenuItem } from "./FolderMenuListItem";
+
+interface Props {
+  startHidden?: boolean;
```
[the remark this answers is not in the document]
StuartHarris: Yeah the name is weird. There are two `<Foldermenu/>`s, one in the nav, which is hidden by default, and the other in the page which is visible by default. Hover is the same for both. Maybe `hiddenUntilHover` is better?

frontend/components/FolderMenu/FolderMenuButton.tsx
StuartHarris: will do

frontend/components/FolderMenu/FolderMenuList.tsx
StuartHarris: 👍

frontend/components/FolderMenu/FolderMenuList.tsx
StuartHarris: The menu drops down in a different place, depending on whether it is in the nav or on the page. Maybe the prop should just be something like `inNav`?

frontend/components/FolderMenu/FolderMenu.tsx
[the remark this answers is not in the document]
frigus02: Ah right. I missed that. Yeah, `hiddenUntilHover` sounds good to me.

frontend/components/FolderMenu/FolderMenuList.tsx
frigus02: Hm yeah, that's tricky. I'd say ideally it's not the FolderMenuList's responsibility to drop down into a different place. The parent should do that. Because the parent also likely knows by how much it needs to shift down.

But since this specifically belongs to the `FolderMenu`, I'm also happy with `inNav` or if we replicate the prop name of the `FolderMenu`, e.g. `hiddenUntilHover`.

frontend/pages/workspaces/[workspaceId]/folders/[folderId]/upload-file.tsx
```diff
@@ -0,0 +1,203 @@
...
+                name="title"
+                onChange={handleCharNumber}
+                id="title"
+                label="Enter file title*"
+                hint="This is the file title as seen by users. Try to be as descriptive as possible. "
+                inputRef={register({
+                  required: true,
+                  maxLength: MAX_CHARS.title,
+                })}
+                aria-invalid={errors.title ? "true" : "false"}
```
Comment (frigus02) [inline frontend/pages/workspaces/[workspaceId]/folders/[folderId]/upload-file.tsx:155]: Hm, do we manually need to set `aria-invalid`? Shouldn't the form submission and `react-hook-form` take care of accessibility concerns?

frontend/pages/workspaces/[workspaceId]/folders/[folderId]/upload-file.tsx
```diff
@@ -0,0 +1,203 @@
...
+                  errors.title &&
+                  `File title is required and cannot be longer than ${MAX_CHARS.title} characters`
+                }
+              />
+              {`${
+                remainingChars.title || MAX_CHARS.title
+              } characters remaining`}
+              {errors.title?.type === "required" && (
+                <p role="alert">This is required</p>
+              )}
```
Comment (frigus02) [inline frontend/pages/workspaces/[workspaceId]/folders/[folderId]/upload-file.tsx:166]: I'm not entirely sure, but I don't think `role=alert` is a good idea here. Basically, I think there should not be more than one `role=alert` on a page at any given time, because screen readers then won't read them out properly.

I think ideally we do the following:
- Add field level errors to input fields as we already do using the `error` prop. This styles them properly and associated them with the input field. So when the user focuses the input fields, it's read out.
- Add an error summary to the top of the form, which either just says "The form has errors" or ideally lists the errors. But even here we don't use `role=alert`. Instead we `focus()` is on form submission. This will also lead to it being read out. And it also resets the focus to the top of the form, so the user can then tab again through the input fields, which will read out their invididual error messages.

That said, the error summary might be a nice to have in the beginning. We could add this in the future.

frigus02 (commented):
Just spotted another thing, sorry.

frontend/pages/workspaces/[workspaceId]/folders/[folderId]/upload-file.tsx
Comment (StuartHarris) [reply frontend/pages/workspaces/[workspaceId]/folders/[folderId]/upload-file.tsx]: re-jigged all this to use `await` and handle errors. Much nicer now :-)

frontend/pages/workspaces/[workspaceId]/folders/[folderId]/upload-file.tsx
Comment (StuartHarris) [reply frontend/pages/workspaces/[workspaceId]/folders/[folderId]/upload-file.tsx:164]: I would have thought so too. But this... https://react-hook-form.com/advanced-usage/

frontend/pages/workspaces/[workspaceId]/folders/[folderId]/upload-file.tsx
Comment (StuartHarris) [reply frontend/pages/workspaces/[workspaceId]/folders/[folderId]/upload-file.tsx:172]: There is now a summary at the bottom, but I'm not giving it focus. Will tallk with Kim about what this should look like, but think we should merge this as is and come back to improve form a11y later, as I think the create folder form could be improved as well.

frontend/components/FolderMenu/FolderMenuList.tsx
StuartHarris: Agree it would be better for the parent to position the menu. But that seems like more work than is necessary right now. Will bear in mind for next time

frigus02: Yeah, agree

frontend/.env.development
```diff
@@ -1,3 +1,2 @@
 ORIGIN=http://localhost:3000
 COOKIE_SECRET=below-particular-sky-mix
-WORKSPACE_SERVICE_GRAPHQL_ENDPOINT=http://localhost:3001/graphql
```
frigus02, line 3: I think we should be able to undo this change now. Without the default environment variable you have to explicitly set it in order to make the frontend work.

frontend/pages/workspaces/[workspaceId]/folders/[folderId]/upload-file.tsx
```diff
@@ -0,0 +1,210 @@
...
+                    message: `File title cannot be longer than ${MAX_CHARS.title} characters`,
+                  },
+                })}
+                aria-invalid={errors.title ? "true" : "false"}
+                error={errors.title?.message}
+              />
+              {`${
+                remainingChars.title || MAX_CHARS.title
+              } characters remaining`}
+              {errors.title?.type === "required" && <p>This is required</p>}
```
Comment (frigus02) [inline frontend/pages/workspaces/[workspaceId]/folders/[folderId]/upload-file.tsx:170]: I feel like this is redundant because the message is already part of the input error prop. See:
image

frigus02:
Long folder titles seem to have regressed in this PR:
image

Before they were multi-line. I think the final design was to make them single line with an ellipsis ("...") at the end. Not sure what's easier to achieve with the current styles.

frigus02 (commented):
Not sure about the regression, but I'm happy to merge otherwise.

frigus02 (approved):

PR #291, opened 2020-09-30
Fix workspace creation

frigus02:
Workspace creation in production doesn’t work at the moment because of various reasons

As far as I can tell it’s:
- Cluster has not enough memory. I’m going to scale it to 2 instances. According to Terraform that should happen automatically. Not sure why it doesn’t. I'm going to fix this in another PR.
- Frontend resource limits are too constraint. It’s getting killed with OOM. This PR increases the limits.
- GraphQL mutation is wrong. It’s using a workspace to create a workspace, but it should be called newWorkspace. This PR updates the schema and mutations.


matt-paul (approved):
🏄‍♂️

frigus02:
Tested on Matt's cluster

PR #293, opened 2020-09-30
Zero downtime

frigus02:
In order to safely update the Kubernetes node pools (see #292), we need to configure our service for zero downtime. This means we need at least:

- More then 1 replica
- A PodDisruptionBudget

This PR adds those things for frontend, workspace-service, ingress.

I also made it so the production cluster has at least 2 nodes. We're reaching the capacity of one node anyway. And for uptime reasons it's better as well.

### DIFF
--- a/frontend/manifests/base/deployment.yaml
+++ b/frontend/manifests/base/deployment.yaml
@@ -3,7 +3,7 @@ kind: Deployment
 metadata:
   name: frontend
 spec:
-  replicas: 1
+  replicas: 2
   selector:
     matchLabels:
       app: frontend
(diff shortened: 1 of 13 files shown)


frigus02:
A pod anti affinity with itself would also be nice to ensure Kubernetes tries to schedule pods of the same deployment on different nodes. But we can do that later:

```yaml
      affinity:
        podAntiAffinity:
          preferredDuringSchedulingIgnoredDuringExecution:
          - labelSelector:
              matchExpressions:
              - key: app
                operator: In
                values:
                - frontend
            topologyKey: "kubernetes.io/hostname"
```

See https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#always-co-located-in-the-same-node

PR #294, opened 2020-09-30
accept dev-$NAME in create-dev-environment.sh

alsuren:


frigus02 (approved):
I wouldn't even mind accepting only `dev-$NAME` and screaming at you when you specify `production`. I think that's what install-everything.sh does.

PR #296, opened 2020-10-01
Update the readme for the frontend following CI cypress work.

edcompton:


frontend/README.md
```diff
@@ -6,21 +6,52 @@ The frontend development server relies on some services that are available in th
...
+WORKSPACE_SERVICE_GRAPHQL_ENDPOINT=http://workspace-service.workspace-service/graphql
+```
+
+Now, when running `yarn dev` the frontend will call the `workspace-service` in your cluster rather than the local stub server.
+
+### Using Azure Active Directory login
+
+By default, when running `yarn dev` your service will mock login, rather than communicating with Azure, in order to reduce dependencies. It is, however, possible to replicate production locally and login normally.
+
+In order to login you will also need to add our shared OAuth 2 client secret to your local environment variables. Ask a coworker for the secret, then create a file called `.env` with the following content:
```
alsuren, line 45: I would prefer it if we told people to pluck it out of their dev cluster using `kubectl get secret -n frontend sessions -o jsonpath={.data.AAD_B2C_CLIENT_SECRET} | base64 -D | fmt`

frontend/README.md
```diff
@@ -6,21 +6,52 @@ The frontend development server relies on some services that are available in th
 
 In order to set up a development cluster, see [infrastructure docs](../infrastructure/README.md). Once your development cluster is set up, you should run this to make the cluster's services available for local development:
```
frigus02, line 7: I think we could move this paragraph and the one above ("The frontend development server relies on some services...") to the "Running against your cluster workspace service" section.

alsuren (approved):

frontend/README.md
edcompton: Good idea 👍

frontend/README.md
```diff
@@ -6,21 +6,52 @@ The frontend development server relies on some services that are available in th
...
 ```
 AAD_B2C_CLIENT_SECRET=<secret>
 ```
 
-Once that is all set up, start the dev server like this:
+Once that is all set up, start the server like this:
 
 ```bash
-yarn dev
+yarn start
```
frigus02, line 54: Running `start` requires you to run `build` before, I think. Maybe we can add that in.
```suggestion
yarn build
yarn start
```

frontend/README.md
```diff
@@ -6,21 +6,52 @@ The frontend development server relies on some services that are available in th
...
+WORKSPACE_SERVICE_GRAPHQL_ENDPOINT=http://workspace-service.workspace-service/graphql
+```
+
+Now, when running `yarn dev` the frontend will call the `workspace-service` in your cluster rather than the local stub server.
+
+### Using Azure Active Directory login
+
+By default, when running `yarn dev` your service will mock login, rather than communicating with Azure, in order to reduce dependencies. It is, however, possible to replicate production locally and login normally.
+
+In order to login you will also need to add our shared OAuth 2 client secret to your local environment variables. Fetch it from your dev cluster with `kubectl get secret -n frontend sessions -o jsonpath={.data.AAD_B2C_CLIENT_SECRET} | base64 -D | fmt`, then create a file called `.env` with the following content:
```
frigus02, line 45: The file should be called `.env.local` because `*.local` files are part of `.gitignore`

frontend/README.md
edcompton: will do

frontend/README.md
edcompton: 👍

frontend/README.md
edcompton: 👍

frigus02 (approved):

PR #300, opened 2020-10-01
delete hello-world service

alsuren:


frontend/server.js
```diff
@@ -86,14 +79,6 @@ async function main() {
 
   const server = express();
 
-  // Dev proxy
-  if (dev) {
-    const { createProxyMiddleware } = require("http-proxy-middleware");
```
frigus02, line 91: Can we remove `http-proxy-middleware` from package.json?

PR #303, opened 2020-10-02
Cleanup IE11 tests

frigus02:
- Clean up `test/` folder, removing the remaining Cypress bits. The folder already doesn't contain any cypress tests anymore. So this just removes the orchestration scripts and Node dependencies.
- Clean up IE Selenium tests a bit, pulling over some fixes I made in #284
- Fix site for IE11. Had some issues with React Hook Form, which needs a special import. Setup ESLint rule to enforce this import.

IE


alsuren (approved):
Looks sensible. Not tried it out though.

frigus02:
I did try it 🙂 

All tests for work locally, except the login test. That makes sense because the login test orchestrates the AAD B2C login, which doesn't exist locally anymore.

PR #304, opened 2020-10-02
Cypress testing for create folder

trcywu:
[Ticket]()

[Zeplin]()

## Acceptance Criteria
Cypress tests for creating folder 

2 tests

1) Fill in title, description, click on submit and new folder is created

2) When on create folder page, click on discard page and go back to previous page

for test 2) cypress was erroring the below

```
Cypress detected a cross origin error happened on page load:
  > Blocked a frame with origin "http://localhost:3000" from accessing a cross-origin frame.
Before the page load, you were bound to the origin policy:
  > http://localhost:3000
A cross origin error happens when your application navigates to a new URL which does not match the origin policy above.
A new URL does not match the origin policy if the 'protocol', 'port' (if specified), and/or 'host' (unless of the same superdomain) are different.
Cypress does not allow you to navigate to a different origin URL within a single test.
You may need to restructure some of your test code to avoid this problem.
Alternatively you can also disable Chrome Web Security in Chromium-based browsers which will turn off this restriction by setting { chromeWebSecurity: false } in ./cypress.json.Learn more
```

The workaround was to turn chromeWebSecurity off in config. More info here https://docs.cypress.io/guides/guides/web-security.html#Disabling-Web-Security

## Pre-review checklist:

- [ ] Tests

- [ ] Documentation

- [ ] Feature flag

## Testing information

- Test plan?

## Test:

- [ ] Tested locally

- Platforms/Browsers:

  - [ ] Internet Explorer 11

  - [ ] Edge

- Accessibility:

  - [ ] Screen Reader

  - [ ] Keyboard Navigation

  - [ ] Magnification

  - [ ] Contrast

  - [ ] Text size 200%

  - [ ] Landscape (Mobile)

### DIFF
--- a/frontend/cypress.json
+++ b/frontend/cypress.json
@@ -7,5 +7,6 @@
   "env": {
     "TEST_WORKSPACE_ID": "750a5fb7-47da-4a8c-8d2e-399ae8d24cde",
     "TEST_WORKSPACE_NAME": "Selenium Testing"
-  }
+  },
+  "chromeWebSecurity": false
 }
--- a/frontend/cypress.json
+++ b/frontend/cypress.json
@@ -7,6 +7,5 @@
   "env": {
     "TEST_WORKSPACE_ID": "750a5fb7-47da-4a8c-8d2e-399ae8d24cde",
     "TEST_WORKSPACE_NAME": "Selenium Testing"
-  },
-  "chromeWebSecurity": false
+  }
 }
--- a/frontend/cypress/integration/frontend/workspaces/folders/create-folder.js
+++ b/frontend/cypress/integration/frontend/workspaces/folders/create-folder.js
@@ -11,12 +11,12 @@ describe("Create folder page", () => {
   });
 
   it("Shows title, form and discard to go back to previous page ", () => {
-    cy.visit(
-      `/workspaces/01bb9a4d-2977-4c43-b28c-2a72b4eda453/folders/create-folder`
-    );
+    const start = `/workspaces/01bb9a4d-2977-4c43-b28c-2a72b4eda453`;
+    cy.visit(start);
+    cy.contains("a", "Create new folder").click();
     cy.contains("h1", "Create a folder");
 
     cy.get("button").contains("Discard").click();
-    cy.go("back");
+    cy.location("pathname").should("eq", start);
   });
 });
--- a/frontend/cypress/integration/frontend/workspaces/folders/create-folder.js
+++ b/frontend/cypress/integration/frontend/workspaces/folders/create-folder.js
@@ -1,0 +1,23 @@
+describe("Create folder page", () => {
+  // Currently only works against Prod or if you manually create a Workspace named "Selenium Testing" in your local/dev cluster environment
+  it("Shows title, renders form and submits", () => {
+    cy.visit(
+      `/workspaces/01bb9a4d-2977-4c43-b28c-2a72b4eda453/folders/create-folder`
+    );
+    cy.contains("h1", "Create a folder");
+
+    cy.get("#title").type("New Folder title");
+    cy.get("#description").type("New Folder description");
+    cy.get("form").submit();
+  });
+
+  it("Shows title, form and discard to go back to previous page ", () => {
+    cy.visit(
+      `/workspaces/01bb9a4d-2977-4c43-b28c-2a72b4eda453/folders/create-folder`
+    );
+    cy.contains("h1", "Create a folder");
+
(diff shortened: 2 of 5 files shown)


frontend/cypress/integration/frontend/workspaces/folders/create-folder.js
```diff
@@ -0,0 +1,23 @@
+describe("Create folder page", () => {
+  // Currently only works against Prod or if you manually create a Workspace named "Selenium Testing" in your local/dev cluster environment
```
frigus02, line 2: It might be obvious, but this also works against the local dev environment with the stub server, doesn't it?

frigus02 (commented):
Looks great. You might want to delete the  `test/.env.local` file. And I want to have a look and understand the cross origin error.

frontend/cypress/integration/frontend/workspaces/folders/create-folder.js
trcywu: yeah it does not that we have stub. Will remove comment

frontend/cypress/integration/frontend/workspaces/folders/create-folder.js
```diff
@@ -0,0 +1,22 @@
...
+  });
+
+  it("Shows title, form and discard to go back to previous page ", () => {
+    cy.visit(
+      `/workspaces/01bb9a4d-2977-4c43-b28c-2a72b4eda453/folders/create-folder`
+    );
+    cy.contains("h1", "Create a folder");
+
+    cy.get("button").contains("Discard").click();
+    cy.go("back");
```
alsuren, line 20: I don't think you mean `cy.go("back")`. 

I would expect a test with this name to:

* visit the workspace homepage (or a folder homepage) so that you have a "previous page" to go back to.
* click on "create new folder"
* (optional) assert that you have navigated to /workspaces/01bb9a4d-2977-4c43-b28c-2a72b4eda453/folders/create-folder
* cy.get("button").contains("Discard").click();
* assert that you have navigated to the previous page that you were on.

frontend/cypress.json
```diff
@@ -7,5 +7,6 @@
   "env": {
     "TEST_WORKSPACE_ID": "750a5fb7-47da-4a8c-8d2e-399ae8d24cde",
     "TEST_WORKSPACE_NAME": "Selenium Testing"
-  }
+  },
+  "chromeWebSecurity": false
```
frigus02, line 11: Because it's a global security settings I'm a bit unhappy about it. I would like to understand why we need it.

The discard button essentially calls `window.history.back()`. I don't understand why would this cause the error. Should we mock/spy on `window.history`? 🤔

frigus02: Ahaaa. If we change the test to this, the web security flag is not necessary:

```js
    const start = `/workspaces/01bb9a4d-2977-4c43-b28c-2a72b4eda453`;
    cy.visit(start);
    cy.contains("a", "Create new folder").click();
    cy.contains("h1", "Create a folder");

    cy.get("button").contains("Discard").click();
    cy.location("pathname").should("eq", start);
```

frigus02: So the cross origin error happens, because we're navigating completely away from the site

trcywu: what does `cy.location("pathname").should("eq", start);` this line do exactly?

trcywu: awesome i'll use this

frigus02: Right, sorry, I didn't explain why this works.

- `cy.visit(...);` starts the browser at the given URL
- Clicking Discard here does a browser back navigation. Since the browser just started this essentially navigates away from the site, which does not work.

This means we can change to:
- `cy.visit(`/workspaces/01bb9a4d-2977-4c43-b28c-2a72b4eda453`)` --> start on workspace homepage
- `cy.contains("a", "Create new folder").click();` --> navigate to folder create page (now we have 1 history entry we can go back to
- `cy.get("button").contains("Discard").click();` --> go back
- `cy.location("pathname").should("eq", start);` --> this gets `window.location.pathname` and checks that it equals the original url we visited

alsuren: so this means we can delete the `"chromeWebSecurity": false` thing now and then ship it?

frigus02: Looks great. You should be able to remove this `"chromeWebSecurity": false`

alsuren (approved):

frigus02 (approved):

PR #318, opened 2020-10-05
Mock Urql client

StuartHarris:
This PR adds a mock Urql client that we can use in our tests. It provides a strongly typed interface for specifying data to be returned during the test run.

### DIFF
--- a/frontend/lib/test-helpers/urql.ts
+++ b/frontend/lib/test-helpers/urql.ts
@@ -1,0 +1,31 @@
+import { DocumentNode } from "graphql";
+import { makeResult, Client, Operation, createClient } from "urql";
+import { fromValue } from "wonka";
+
+import { Query } from "../generated/graphql";
+
+export type MockUrqlClient = (
+  queries: Array<[DocumentNode, Partial<Query>]>
+) => Client;
+
+export const mockUrqlClient: MockUrqlClient = (queries) => {
+  const client = createClient({ url: "http://localhost", exchanges: [] });
+  return Object.assign(client, {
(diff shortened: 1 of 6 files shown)


frontend/lib/test-helpers/urql.ts
```diff
@@ -0,0 +1,31 @@
+import { DocumentNode } from "graphql";
+import { makeResult, Client, Operation, createClient } from "urql";
+import { fromValue } from "wonka";
+
+import { Query } from "../generated/graphql";
+
+export type MockUrqlClient = (
+  queries: Array<[DocumentNode, Partial<Query>]>
```
frigus02, line 8: Nice. I think the only downside is that you need to specify slightly more fields now with the mocks. E.g. the GetWorkspacesQuery actually only returns `id` and `title`, not the `description`. But the `Query` type doesn't know that so you need to specify `description` as well.

But the only thing that binds the actual return type to the query are the `use*` functions. So the only way to make the type here more accurate is if we tried to extract the info from them, which might be overkill.

frigus02 (approved):

frontend/lib/test-helpers/urql.ts
frigus02: Okay, I don't think using the `use*` functions to bind query and response types together actually works.

StuartHarris: Yeah good point. I don't think my TypeScript is good enough for that :-), but if you fancy having a stab....  In the meantime, I'll merge this anyway as I think it is marginally better than before (and gets rid of the typescript error that the old pattern had).

PR #321, opened 2020-10-06
Generate GraphQL schema in workspace service

frigus02:
At the moment we have to keep the `frontend/schema.graphql` file up-to-date manually. This is annoying and potentially prone to errors.

This PR changes the process to:
- Have the workspace-service generate a GraphQL schema file and ensure using an integration test, that the schema is up-to-date.
- The frontend generates it's definitions based on this schema file.


workspace-service/src/graphql/schema.rs
```diff
@@ -0,0 +1,112 @@
...
+    Subscription: SubscriptionType + Send + Sync + 'static,
+{
+    let query = include_str!("./introspection_query.graphql");
+    let result = QueryBuilder::new(query)
+        .operation_name("IntrospectionQuery")
+        .execute(schema)
+        .await?;
+    let mut data = result.data;
+    sort_value(&mut data);
+    Ok(format!("{:#}", data))
```
alsuren, line 71: This sorting definitely feels like it should be done in the upstream library (I assume it's because they're using hashmaps under the hood somewhere?). Probably worth reporting, and linking to this PR as a work-around.

workspace-service/Makefile
```diff
@@ -64,6 +64,10 @@ migrate: ## Run migrations on current cluster
 prepare: migrate-local ## Prepares new version of sqlx-data.json
 	cargo clean --package workspace_service && DATABASE_URL=$(localdb) cargo sqlx prepare -- --bin workspace_service
 
+.PHONY: graphql-schema.json
```
StuartHarris, line 67: Just thinking out loud, but if the target wasn't `.PHONY` and it depended on `src/graphql`, then make would only run the generate if the graphql code had been updated?

StuartHarris (commented):
I think this is amazing! And it opens the door to us potentially using a reduced version of the json graphql schema in the Urql client

workspace-service/src/graphql/schema.rs
frigus02: Yeah agree. And that's actually a good point. I'm going to open an issue there and see what they say.

frontend/stubServer/index.js
```diff
@@ -3,20 +3,21 @@ const path = require("path");
...
 const { ApolloServer, gql } = require("apollo-server-express");
 const app = require("express")();
+const { buildClientSchema, printSchema } = require("graphql/utilities");
 
-const resolvers = require("./resolver");
+const createResolvers = require("./resolver");
 
 const port = 3001;
-const schemaCode = fs.readFileSync(
-  path.join(__dirname, "schema", "../../schema.graphql"),
```
alsuren, line 11: hrm. I actually quite like having schema.graphql in schema definition language format.

workspace-service/Makefile
frigus02: I would need to tell `make` about the dependencies for the file then. They are basically all the Rust source files involved in the GraphQL schema plus the introspection query. It felt easier to make it phony. 🙈

frontend/stubServer/index.js
frigus02: I was wondering if anyone is looking at the schema. Sadly I cannot find a Rust library to generate a GraphQL schema file based on the introspection JSON result.

With the Node.js `graphql` module, it's super easy to convert between those 2 representations. That's basically the other option here. We could generate the schema.graphql file using a bit of JavaScript. Either by starting the workspace service and calling it, or by basing it on the JSON file.

alsuren (approved):
I think this puts us in a better place than we are currently. I'm happy to iterate on the rest and potentially simplify things at a later date.

workspace-service/Makefile
StuartHarris: Fair enough. Isn't it just this though?
```make
graphql-schema.json: src/graphql
```

frigus02: Ah, hm. Yeah, I think you're right. That looks good. Let me change that

workspace-service/src/graphql/schema.rs
frigus02: Okay. I just reproduced this with a simple example: https://github.com/frigus02/test-async-graphql-deterministic-introspection

I then updated to the latest alpha version (2.0.0-alpha.23) of `async-graphql`. In this version I cannot reproduce the problem anymore. Seems to be fixed.

This is great. I'm just going to add a comment here to remove the custom sort after the the upgrade.

workspace-service/Makefile
frigus02: It works, obviously. I don't know what I was thinking yesterday.

workspace-service/src/graphql/schema.rs
frigus02: Added a comment explaining why sorting is necessary and when we can remove it.

PR #322, opened 2020-10-06
Adds missing tests for file upload

StuartHarris:
Note: Changes cypress tests to run with `NODE_ENV=test`. This allows it to be independent of your development environment, so you don't also _have_ to use the stub server for development.

Adds a missing test to generate a snapshot of the file upload page.
Adds a couple of cypress tests (note we can't intercept calls to external URLs like Azure Blob Storage, so one test checks that the error is correct).


frontend/server.js
```diff
@@ -57,14 +57,14 @@ const runSessionDbMigration = async (pool) => {
 };
 
 async function main() {
-  const dev = process.env.NODE_ENV !== "production";
-  const dotEnvFileSegment = dev ? "development" : "production";
+  let env = process.env.NODE_ENV || "development";
```
frigus02, line 60: I think that can be `const`

```suggestion
  const env = process.env.NODE_ENV || "development";
```

frigus02 (approved):
Makes sense to me 👍

frontend/server.js
StuartHarris: Ha! yeah, left over from refactor. Thanks

PR #330, opened 2020-10-07
Update method to set env var in GitHub Actions

frigus02:
The old method (writing ::set-env to stdout) had a security
vulnerability. See: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

GitHub actions now provides a new way of setting environment variables,
that can be used by later steps in the workflow. It works by writing the
desired variable to a temporary file. This updates all our workflows to
use this new method.


StuartHarris (approved):

frigus02:
The remaining warnings come from the Docker login, Cypress and Terraform setup actions.

PR #333, opened 2020-10-07
AB2202 move file to destination container

StuartHarris:
## Move uploaded file from `upload` container to `files` container

- [x] provision files container with Terraform
- [x] move file to destination synchronously (only works for files <256MB, which we think is good for now)

At some point we may introduce an async copy for files larger than 256MB.

Eventually the [BlobIndex](https://docs.microsoft.com/en-gb/azure/storage/blobs/storage-manage-find-blobs?tabs=azure-portal#regional-availability-and-storage-account-support) feature will come out of preview and we will be able to tag files as having a validating status so we don't have to move them between containers.

[Ticket](https://dev.azure.com/futurenhs/FutureNHS/_workitems/edit/2202)

## Pre-review checklist:

- [x] Tests

- [ ] Documentation

- [ ] Feature flag

## Testing information

- Test plan?

## Test:

- [x] Tested locally

- Platforms/Browsers:

  - [ ] Internet Explorer 11

  - [ ] Edge

- Accessibility:

  - [ ] Screen Reader

  - [ ] Keyboard Navigation

  - [ ] Magnification

  - [ ] Contrast

  - [ ] Text size 200%

  - [ ] Landscape (Mobile)


workspace-service/src/azure/blob.rs
```diff
@@ -0,0 +1,58 @@
...
+pub async fn copy_blob_from_url(url: &Url, azure_config: &super::Config) -> Result<String> {
+    let target_storage_account = azure_config
+        .files_container_url
+        .host()
+        .expect("invalid files_container_url")
+        .to_string();
+    let target_storage_account = target_storage_account
+        .split('.')
+        .next()
+        .expect("invalid files_container_url");
```
frigus02, line 17: I'm trying to get this working with a local Azure storage emulator (Azurite). This uses a different scheme, where the container name is part of the path, not the hostname. Ideally we would support that. BUT: the Azure SDK for Rust make the same assumption, so it's probably fine.

workspace-service/src/azure/blob.rs
```diff
@@ -0,0 +1,58 @@
...
+        .path_segments()
+        .expect("invalid files_container_url")
+        .next()
+        .expect("cannot get container name from url");
+
+    let target_blob = url
+        .path_segments()
+        .ok_or_else(|| anyhow!("invalid temporary_blob_storage_path"))?
+        .nth(1)
+        .ok_or_else(|| anyhow!("cannot get blob name from temporary_blob_storage_path"))?;
```
frigus02, line 30: I think it would be good to add some validation here. Basically I want to make sure the file comes from the uploads container, not the files container (or any other future container our access key works for). I think you could otherwise cheat your way into having access to another file knowing it's blob storage path.

frigus02 (commented):
Looks great. I do have 1 security concern, though.

workspace-service/src/config.rs
```diff
@@ -53,5 +53,14 @@ pub struct Config {
...
+    pub upload_container_url: Option<Url>,
+
+    /// The Azure Blob Storage Container URL for files
+    #[structopt(
+        long,
+        env = "FILES_CONTAINER_URL",
+        parse(try_from_str = str::parse),
+        required_unless_one(&["selfcheck-only", "generate-schema-only"])
+    )]
+    pub files_container_url: Option<Url>,
```
alsuren, line 65: (something for a separate PR): all of these required_unless_one() fields mean that we've got a bunch of optional things in our config that we need to unwrap later. I would feel much happier if we could make them required, but use subcommands for them instead, like:
```
cargo run -- selfcheck
```
(which needs none of the options), and:
```
cargo run -- serve --files-container-url=xyz, ...
```
which needs all of the options, and doesn't use Option for them

workspace-service/src/azure/blob.rs
StuartHarris: you mean the storage account is part of the path?

workspace-service/src/graphql/files.rs
```diff
@@ -85,43 +86,30 @@ impl FilesMutation {
...
+            &destination,
             pool,
         )
-        .await
-    }
-}
+        .await?
+        .into();
 
-async fn create_file(
```
alsuren, line 103: (for context, this was split out at the top level so that it would be easier to write a test, but then we decided that it wasn't worth testing until it did something more interesting ;-) )

workspace-service/src/azure/blob.rs
frigus02: Yes. The path is `http://localhost:10000/storageaccountname/container/blobname`

workspace-service/src/azure/blob.rs
StuartHarris: ahh good point

workspace-service/src/azure/blob.rs
frigus02: But like I said, Azure SDK for Rust makes the same assumption. So it's probably fine for us to do the same. Because we need a workaround anyway if we want to use Azurite locally.

workspace-service/src/azure/blob.rs
```diff
@@ -0,0 +1,58 @@
...
+        .split('.')
+        .next()
+        .expect("invalid files_container_url");
+
+    let target_container = azure_config
+        .files_container_url
+        .path_segments()
+        .expect("invalid files_container_url")
+        .next()
+        .expect("cannot get container name from url");
```
alsuren, line 24: This top block feels like command-line argument parsing. Could it be shifted to startup-time, or could we have some helpers for blob storage url parsing, so that this function is easier to read?

alsuren (approved):
shape looks good. Only comments.

cobwebfx (approved):

workspace-service/src/azure/blob.rs
StuartHarris: added an `From` impl , but we can deffo move to startup. Will do that in another PR

workspace-service/src/azure/blob.rs
```diff
@@ -3,42 +3,79 @@ use async_compat::Compat;
...
-        .nth(1)
-        .ok_or_else(|| anyhow!("cannot get blob name from temporary_blob_storage_path"))?;
+    let input: FileParts = url.clone().try_into()?;
+    let target: FileParts = azure_config.files_container_url.clone().try_into()?;
+    let source: FileParts = azure_config.upload_container_url.clone().try_into()?;
+
+    if input.account != source.account {
+        return Err(anyhow!(
+            "source file is from an unsupported storage account"
+        ));
```
alsuren, line 58: (optional) `anyhow::bail!("source file is from an unsupported storage account")`

workspace-service/src/azure/blob.rs
```diff
@@ -0,0 +1,125 @@
...
+struct FileParts {
+    account: String,
+    container: String,
+    blob: Option<String>,
+}
+
+impl TryFrom<Url> for FileParts {
+    type Error = anyhow::Error;
+
+    fn try_from(value: Url) -> Result<Self, Self::Error> {
```
frigus02, line 19: You could possibly implement this for `&Url` instead, which would avoid an additional clone for on the calling side.

alsuren (approved):

frigus02 (approved):

workspace-service/src/azure/blob.rs
frigus02: Nice. That's much easier to understand 👍

workspace-service/src/azure/blob.rs
StuartHarris: Oh, I like that!

workspace-service/src/azure/blob.rs
StuartHarris: Yeah I tried that, but it then doesn't match the trait signature :-(

StuartHarris: Oh actually ignore me, that was probably as I didn't change the `impl` line!

StuartHarris: done

PR #335, opened 2020-10-08
cypress tests for files

cobwebfx:
[Ticket] 2198

## Pre-review checklist:

- [x] Tests

## Test:

- [x] Tested locally

- Platforms/Browsers:

### DIFF
--- a/frontend/cypress/integration/frontend/workspaces/folders/browse-files.js
+++ b/frontend/cypress/integration/frontend/workspaces/folders/browse-files.js
@@ -1,0 +1,112 @@
+describe("Browse files", () => {
+  it("Shows title and summary", () => {
+    const start = `/workspaces/750a5fb7-47da-4a8c-8d2e-399ae8d24cde/folders/b337907b-439d-4d05-9937-4975fa00bed7`;
+    cy.visit(start);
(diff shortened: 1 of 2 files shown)


frigus02 (approved):

frontend/cypress/integration/frontend/workspaces/folders/browse-files.js
```diff
@@ -0,0 +1,112 @@
...
+      cy.viewport(1200, 1080);
+    });
+
+    it("Shows files on desktop", () => {
+      const start = `/workspaces/750a5fb7-47da-4a8c-8d2e-399ae8d24cde/folders/b337907b-439d-4d05-9937-4975fa00bed7`;
+      cy.visit(start);
+      cy.get("tr")
+        .eq(1)
+        .within(() => {
+          cy.get("td").eq(0).get(`svg[width="28"]`);
```
declanslevin, line 21: It doesn't look like anything is being done with this element, do we need this get() call?

frigus02: It just asserts that the element with this selector exists. Would be nice if we could verify the type of SVG here, but there is not easy way for that at the moment.

PR #347, opened 2020-10-09
Forward x-user-auth-id header to workspace-service

alsuren:
On login:

* take azure auth id and call getOrCreateUser mutation
* store more things in the User object in the frontend (we renamed .id to authId and added .id as the id of user in the workspace-service database)

We also refactored requireEnv so that it could be used in js land. We also hand-rolled our getOrCreateUser query using urql rather than generating it, because typescript :(.

Would be interesting to hear people's ideas about how to make requireEnv less terrible.

### DIFF
--- a/frontend/pages/api/health.ts
+++ b/frontend/pages/api/health.ts
@@ -4,6 +4,7 @@ const requiredEnvVars = [
   "EVENTGRID_TOPIC_ENDPOINT",
   "EVENTGRID_TOPIC_KEY",
   "PG_URL",
+  "WORKSPACE_SERVICE_GRAPHQL_ENDPOINT",
 ];
 
 export default (_: NextApiRequest, res: NextApiResponse) => {
--- a/frontend/server.js
+++ b/frontend/server.js
@@ -121,7 +121,7 @@ async function main() {
           },
           (profile, done) => {
             done(null, {
-              id: profile.sub,
+              authId: profile.sub,
               name: profile.displayName,
               emails: profile.emails,
             });
--- a/frontend/server.js
+++ b/frontend/server.js
@@ -12,15 +12,7 @@ const Noop = require("./lib/server/noop-passport-strategy");
 const url = require("url");
 const { promises: fs } = require("fs");
 const { getOrCreateUser } = require("./lib/server/graphql");
-
-const requireEnv = (name) => {
-  const value = process.env[name];
-  if (!value) {
-    throw new Error(`Environment variable ${name} is required`);
-  }
-
-  return value;
-};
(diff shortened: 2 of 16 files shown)


frontend/pages/api/health.ts
```diff
@@ -4,6 +4,7 @@ const requiredEnvVars = [
   "EVENTGRID_TOPIC_ENDPOINT",
   "EVENTGRID_TOPIC_KEY",
   "PG_URL",
+  "WORKSPACE_SERVICE_GRAPHQL_ENDPOINT",
```
alsuren, line 7: This is because our `requireEnv("WORKSPACE_SERVICE_GRAPHQL_ENDPOINT")` call is lazy (it has to be, because we don't do .env parsing until after import time. Can we think of a better way to make all of this a bit more lined up?

frontend/server.js
```diff
@@ -119,10 +112,22 @@ async function main() {
             allowHttpForRedirectUrl: dev,
             isB2C: true,
           },
-          (profile, done) => {
-            done(null, {
-              id: profile.sub,
+          async (profile, done) => {
+            const response = await getOrCreateUser({
```
alsuren, line 116: If this fails, can we propagate the error up to passport?

edcompton: Added `try/catch` to bubble up errors encountered by either the local or OIDC (Open ID Connect) Passport Strategies.

frigus02 (approved):

frontend/pages/api/health.ts
frigus02: Would be great to line this up. Maybe using types? Something like:

```ts
export const ENV_VAR_NAMES = [
  "EVENTGRID_TOPIC_ENDPOINT",
  "EVENTGRID_TOPIC_KEY",
  ...
] as const;
export type EnvVarName = typeof ENV_VAR_NAMES[number];
```

Where `api/health.ts` expects all of them to be valid and the `requireEnv` function expects `EnvVarName` as the variable name instead of `string`. Then we forbit usage of `process.env` using ESLint. Et voila (is that correct french?)

alsuren: I think we're on the same page here. I will spin it out as a separate PR.

frigus02: Nice. Yeah, definitely separate PR.

alsuren:
The force-push is a rebase that results in a file tree that's identical to something that's already passed, so I'm not waiting for CI. 🤠

frontend/pages/api/health.ts
alsuren: https://github.com/FutureNHS/futurenhs-platform/pull/360

PR #358, opened 2020-10-12
Ab1796 download file be

cobwebfx:
[AB#1796](https://dev.azure.com/futurenhs/75407963-f812-43e5-a734-1059bbc036a3/_workitems/edit/1796)

## Acceptance Criteria
- [x] File downloads on click of "Download file"


workspace-service/src/graphql/files.rs
```diff
@@ -19,8 +21,10 @@ pub struct File {
     pub file_name: String,
     #[field(desc = "The type of the file")]
     pub file_type: String,
-    #[field(desc = "The blob storage path for the file")]
-    pub blob_storage_path: String,
+    // #[field(desc = "The blob storage path for the file")]
+    // pub blob_storage_path: String,
```
frigus02, line 25: We can remove that commented code.

workspace-service/src/graphql/files.rs
```diff
@@ -39,20 +43,25 @@ pub struct NewFile {
...
+    fn from_db(d: db::File, c: &Config) -> Result<Self> {
+        Ok(Self {
             id: d.id.into(),
             title: d.title,
             description: d.description,
             folder: d.folder.into(),
             file_name: d.file_name,
             file_type: d.file_type,
-            blob_storage_path: d.blob_storage_path,
+            // blob_storage_path: d.blob_storage_path,
```
frigus02, line 55: We can remove that commented code.

frontend/.env.development
```diff
@@ -1,3 +1,4 @@
 ORIGIN=http://localhost:3000
 COOKIE_SECRET=below-particular-sky-mix
-WORKSPACE_SERVICE_GRAPHQL_ENDPOINT=http://localhost:3001/graphql
+#WORKSPACE_SERVICE_GRAPHQL_ENDPOINT=http://localhost:3001/graphql
+WORKSPACE_SERVICE_GRAPHQL_ENDPOINT=http://localhost:3030/graphql
```
frigus02, line 4: We should probably revert that change.

frontend/server.js
```diff
@@ -174,6 +174,19 @@ async function main() {
...
 
+  server.get("/workspaces/:workspaceId/download/:fileId", async (req, res) => {
+    try {
+      const response = await getFileDownloadUrl({
+        fileId: req.params.fileId,
+      });
+      const { temporaryBlobStoragePath } = response.file;
+      res.redirect(307, temporaryBlobStoragePath);
+    } catch (err) {
+      console.log(err);
```
frigus02, line 185: Do we want to leave the console log in there should should we remove it?

frigus02 (commented):
Looks great

workspace-service/src/graphql/files.rs
```diff
@@ -39,20 +43,25 @@ pub struct NewFile {
     pub temporary_blob_storage_path: String,
 }
 
-impl From<db::File> for File {
-    fn from(d: db::File) -> Self {
-        Self {
+impl File {
+    fn from_db(d: db::File, c: &Config) -> Result<Self> {
+        Ok(Self {
```
alsuren, line 48: (optional) could be more like:
```
impl TryFrom<db::File> for File {
    type Error = anyhow::Error;

    fn try_from(value: db::File) -> Result<Self, Self::Error> {
```
like it's done in workspace-service/src/azure/blob.rs.

frigus02: The problem is that we need a reference to the Azure `Config`. `From` and `TryFrom` don't allow a second parameter. We could implement `TryFrom` for something like `(File, &Config)`, but I'm not sure that's better.

I do wonder if there is a pattern for that stuff, though. Would that be a builder pattern then?

workspace-service/src/graphql/files.rs
```diff
@@ -39,20 +43,25 @@ pub struct NewFile {
...
             description: d.description,
             folder: d.folder.into(),
             file_name: d.file_name,
             file_type: d.file_type,
-            blob_storage_path: d.blob_storage_path,
+            // blob_storage_path: d.blob_storage_path,
+            temporary_blob_storage_path: azure::create_download_sas(
+                c,
+                &d.blob_storage_path.parse()?,
+            )?
```
alsuren, line 59: temporary_blob_storage_path probably wants to be called temporary_download_url or something?

workspace-service/src/graphql/files.rs
alsuren: ... actually, we're eventually going to want to emit an event whenever anyone clicks on a download link, but not when they just list the files. I'm wondering if we want a separate DownloadLink type or something, to make the distinction more explicit.

workspace-service/src/graphql/files.rs
frigus02: Yeah, we made it match the property of `NewFile`, but I think I agree. The caller doesn't need to know that it's a blob storage path, so `temporary_download_url` is nicer.

workspace-service/src/graphql/files.rs
frigus02: Yeah, I am stuggeling a bit. On the one hand I think it's nice to make this part of `File`. We could optimize the code to only actually generate a download URL if the request asks for the field. That way we would also know when to send the event.

However it could be clearer if we essentially "create" a download. Basically I wouldn't even mind this being a mutation, e.g. "create_file_download_url".

I can't really decide what's better here, to be honest.

alsuren: I think a separate query/mutation will make things clearer. I'm still not sure when to call something a query and when to call it a mutation. I guess if it's a mutation then it's less likely to be cached anywhere.

frontend/.env.development
alsuren: (this can live in frontend/.env.local or frontend/.env.development.local if you want, and it won't be checked in).

Stu was suggesting that we should make `yarn cypress:local` use a `test` environment, with its own `.env.test` and `.env.test.local`. Sounds like a good idea to me.

frontend/server.js
```diff
@@ -174,6 +174,19 @@ async function main() {
     res.redirect("/");
   });
 
+  server.get("/workspaces/:workspaceId/download/:fileId", async (req, res) => {
```
alsuren, line 177: Is there any way we could use the nextjs router for this? (make a `pages/workspaces/[workspaceId]/download/[fileId].ts` or `pages/api/...`) and generate the interfaces + types? frontend/lib/server/graphql.js is a bit of an untyped minefield and I feel sad about making it.

workspace-service/src/graphql/files.rs
frigus02: > I'm still not sure when to call something a query and when to call it a mutation
Same. I kinda view queries like an HTTP `GET` request, which should not causes any noticable side effects and may be cachable. Not sure that's always correct.

I think I'd be happy to change that. @cobwebfx do you want to give this a shot?

frontend/server.js
frigus02: We could put it inside `pages/api/` but then the URL has `api/` inside it. I don't really want to make URLs worse just because Next.js' router want this.

Maybe we do have to start converting server.js to TypeScript at some point 😐

workspace-service/src/graphql/files.rs
cobwebfx: Sure

frigus02 (approved):

PR #362, opened 2020-10-13
Backend permission check

alsuren:
[Ticket]()

[Zeplin]()

## Pre-review checklist:

- [x] Tests

- [x] Documentation

- [ ] ~~Analytics (user analytics, e.g. Google Analytics)~~

- [ ] ~~Feature flag~~

## Testing information

no

- Test plan?

Submit a request to `createWorkspace` without a valid `auth_id` (e.g. one that isn't a platform admin) and the call should fail. 

## Test:

- [x] Tested locally

### DIFF
--- a/workspace-service/src/db.rs
+++ b/workspace-service/src/db.rs
@@ -301,6 +301,7 @@ pub struct User {
     pub is_platform_admin: bool,
 }
 
+#[cfg(not(test))]
 impl User {
     pub async fn find_by_auth_id(auth_id: &Uuid, pool: &PgPool) -> Result<Option<User>> {
         let user = sqlx::query_file_as!(User, "sql/users/find_by_auth_id.sql", auth_id)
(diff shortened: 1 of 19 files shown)


workspace-service/src/graphql/workspaces.rs
```diff
@@ -75,11 +75,18 @@ impl WorkspacesMutation {
     ) -> FieldResult<Workspace> {
         let pool = context.data()?;
         let event_client: &EventClient = context.data()?;
+        let auth_id = context
+            .data::<super::IncomingHeaders>()?
+            .auth_id
+            .clone()
+            .ok_or_else(|| anyhow::anyhow!("no x-user-auth-id header"))?;
```
alsuren, line 82: Let's call this `context.data::<RequestingUser>()?.auth_id` and let's not make it Option. If RequestingUser is not on context then it .data() can just fail. We can probably also get away without the clone if we're clever about it.

workspace-service/src/db.rs
```diff
@@ -302,6 +302,13 @@ pub struct User {
 }
 
 impl User {
+    pub async fn find_by_auth_id(auth_id: &Uuid, pool: &PgPool) -> Result<Option<User>> {
+        let user = sqlx::query_file_as!(User, "sql/users/find_by_auth_id.sql", auth_id)
+            .fetch_optional(pool)
```
alsuren, line 307: let's let this function fail if it can't find the user: Result<Option<User>>  is overkill.

edcompton (approved):

PR #366, opened 2020-10-14
Setup script fixes

alsuren:
[AB#1817](https://dev.azure.com/futurenhs/75407963-f812-43e5-a734-1059bbc036a3/_workitems/edit/1817)

## Acceptance Criteria

Now that create-workspace requires you to be admin:

* Make an admin user as part of the initial migrations (I think this is the best way to bootstrap the cluster)
* Make the create-folder requests using that admin user's auth_id.

## Also includes:

* create-file-if-needed.sh is broken, so we commented it out.

## Pre-review checklist:

- [x] Tests

## Test:

- [x] Tested locally

### DIFF
--- a/workspace-service/migrations/20201014134400_create_admin_user.sql
+++ b/workspace-service/migrations/20201014134400_create_admin_user.sql
@@ -1,0 +1,9 @@
+INSERT into users (
+    auth_id,
+    name,
+    is_platform_admin
+) VALUES (
+    'feedface-0000-0000-0000-000000000000',
+    'Initial Admin User',
+    true
+);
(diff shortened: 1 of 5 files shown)


workspace-service/migrations/20201014134400_create_admin_user.sql
```diff
@@ -0,0 +1,9 @@
+INSERT into users (
+    auth_id,
+    name,
+    is_platform_admin
+) VALUES (
+    'feedface-0000-0000-0000-000000000000',
```
frigus02, line 6: Is feedface a reference so something? Anyway, it made me find https://en.wikipedia.org/wiki/Hexspeak, which is great 😄

frigus02 (approved):

workspace-service/migrations/20201014134400_create_admin_user.sql
alsuren: It is indeed great, and also where I found that name. We have added a deadbeef non-admin user in our next PR as well.

PR #368, opened 2020-10-14
AB#1797 - 'Delete File'

matt-paul:
[Ticket](https://dev.azure.com/futurenhs/FutureNHS/_boards/board/t/FutureNHS%20Team/Stories/?workitem=1797)

[Figma](https://www.figma.com/file/tYFsqhG61CleWS1tUapAar/Source-of-Truth-UI?node-id=467%3A0)

## Acceptance Criteria
- [x] Soft delete when hitting Graphql deleteFile()

This PR contains the backend work; following PR will make use of this in the frontend.

PR #364, opened 2020-10-14
AB#1797 - Refactor Menu

matt-paul:
[Ticket](https://dev.azure.com/futurenhs/FutureNHS/_boards/board/t/FutureNHS%20Team/Stories/?workitem=1797)

[Figma](https://www.figma.com/file/tYFsqhG61CleWS1tUapAar/Source-of-Truth-UI?node-id=467%3A0)

## Acceptance Criteria

This PR refactors the existing FolderMenu component, so that it can contain other menu items depending on where it is used.

There seemed to be two choices here: Have separate components that contain there own menu items and import, or have a single component that is given its items at the place of use.  I went with the second option because I like to have the items easily visible on the page and don't feel that the pages are too crowded, but i appreciate this is subjective, open to discussion if there are other factors that I haven't considered.

## Pre-review checklist:

- [x] Tests

- [ ] Documentation

- [ ] Analytics (user analytics, e.g. Google Analytics)

- [ ] Observability (metrics/tracing)

- [ ] Feature flag

- [ ] Data columns adhere to the [Dublin Core Metatdata schema](https://www.dublincore.org/specifications/dublin-core/dcmi-terms/). Further information is available [here](https://www.gov.uk/government/publications/recommended-open-standards-for-government/using-metadata-to-describe-data-shared-within-government)

## Testing information

- Is there any special setup required?

- Test plan?

## Test:

- [ ] Tested locally

- Platforms/Browsers:

  - [ ] Google Chrome (latest version)

  - [ ] Internet Explorer 11

  - [ ] Edge

- Accessibility:

  - [ ] Screen Reader

  - [ ] Keyboard Navigation

  - [ ] Magnification

  - [ ] Contrast

  - [ ] Text size 200%

  - [ ] Touch target areas (Mobile)

  - [ ] Landscape (Mobile)

PR #372, opened 2020-10-15
AB#1797 - Delete File - Tie together frontend and backend

cobwebfx:
[Ticket](https://dev.azure.com/futurenhs/FutureNHS/_boards/board/t/FutureNHS%20Team/Stories/?workitem=1797)

[Figma](https://www.figma.com/file/tYFsqhG61CleWS1tUapAar/Source-of-Truth-UI?node-id=467%3A0)

This ties together the [front end](https://github.com/FutureNHS/futurenhs-platform/pull/364) and [backend](https://github.com/FutureNHS/futurenhs-platform/pull/368/files) for the 'delete file' ticket.  Cypress tests will come in the next PR.

Screenshot 2020-10-19 at 20 53 05

Also included, [URQL dev tools](https://github.com/FormidableLabs/urql-devtools)..
Screenshot 2020-10-19 at 20 55 28
Screenshot 2020-10-19 at 20 54 47

## Acceptance Criteria
Application soft deletes file

### DIFF
--- a/frontend/components/Menu/MenuListItem.tsx
+++ b/frontend/components/Menu/MenuListItem.tsx
@@ -7,14 +7,26 @@ interface ListItemProps {
   className?: string;
   href: string;
   children: ReactNode;
+  isButton?: boolean;
+  onClick?: () => void;
 }
 
-const ListItem = ({ className, href, children }: ListItemProps) => {
+const ListItem = ({
+  className,
+  href,
+  children,
+  isButton,
+  onClick,
+}: ListItemProps) => {
   return (
     <li className={className}>
-      <Link href={href} passHref>
-        <a>{children}</a>
-      </Link>
+      {isButton ? (
+        <button onClick={onClick}>{children}</button>
+      ) : (
+        <Link href={href} passHref>
+          <a>{children}</a>
+        </Link>
+      )}
     </li>
   );
 };
@@ -61,11 +73,15 @@ export type MenuItem = {
   title: string;
   icon: ReactNode;
   href: string;
+  isButton?: boolean;
+  onClick?: () => void;
 };
 
 interface MenuListItemProps extends ComponentPropsWithoutRef<"li"> {
   title: string;
   href: string;
+  isButton?: boolean;
+  onClick?: () => void;
 }
 
 const MenuListItem: FC<MenuListItemProps> = ({ title, children, ...props }) => {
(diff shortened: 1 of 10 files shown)


frontend/pages/workspaces/[workspaceId]/folders/[folderId]/files/[fileId].tsx
```diff
@@ -62,6 +63,17 @@ const FileHomepage = () => {
     variables: { id: fileId },
   });
 
+  const [, deleteFile] = useDeleteFileMutation();
+
+  const onClick = async () => {
+    const message = "Are you sure you want to delete this file?";
+    const result = window.confirm(message);
+    if (result) {
+      await deleteFile({ id: fileId as string });
```
Comment (frigus02) [inline frontend/pages/workspaces/[workspaceId]/folders/[folderId]/files/[fileId].tsx:72]: Above it looks like `fileId` is already a `string`. Why do we need the explicit cast?

frontend/components/Menu/MenuListItem.tsx
```diff
@@ -7,14 +7,26 @@ interface ListItemProps {
   className?: string;
   href: string;
   children: ReactNode;
+  isButton?: boolean;
+  onClick?: () => void;
```
frigus02, line 11: If possible, I'd prefer a type, which let's you pass in either an `onClick` handler or an `href`, but not both. Would something like this work?

```ts
interface LinkHandler {
  type: "link";
  href: string;
}

interface ButtonHandler {
  type: "button";
  onClick: () => void;
}

interface ListItemProps {
  className?: string;
  children: ReactNode;
  handler: LinkHandler | ButtonHandler;
}
```

frontend/pages/workspaces/[workspaceId]/folders/[folderId]/files/[fileId].tsx
Comment (matt-paul) [reply frontend/pages/workspaces/[workspaceId]/folders/[folderId]/files/[fileId].tsx:72]: Screenshot 2020-10-19 at 12 51 10

Comment (frigus02) [reply frontend/pages/workspaces/[workspaceId]/folders/[folderId]/files/[fileId].tsx:72]: Is that because `router.query` has this type? Hm. I think then let's check for that. Maybe we replace this line above with a check that covers all cases:

```diff
-  fileId = (fileId || "unknown").toString();
+  if (fileId === undefined || Array.isArray(fileId)) {
+    throw new Error("fileId required in URL");
+  }
```

frigus02 (approved):

frigus02 (approved):

PR #386, opened 2020-10-19
Invalidate workspace service cache

frigus02:
sqlx prepare seems to have some cache issues. Sometimes it just won't
produce the correct result and the only way to fix it is to clear the
cache.

This may have been fixed in the latest master branch of sqlx. We should
try to update.

### DIFF
--- a/.github/workflows/workspace-service-branch.yaml
+++ b/.github/workflows/workspace-service-branch.yaml
@@ -59,7 +59,7 @@ jobs:
             ~/.cargo/registry
             ~/.cargo/git
             futurenhs-platform/workspace-service/target
-          key: cargo-${{ hashFiles('futurenhs-platform/workspace-service/Cargo.lock') }}-a_
+          key: cargo-${{ hashFiles('futurenhs-platform/workspace-service/Cargo.lock') }}-1
       - name: Build
         working-directory: futurenhs-platform/workspace-service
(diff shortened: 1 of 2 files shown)


.github/workflows/workspace-service-branch.yaml
```diff
@@ -59,7 +59,7 @@ jobs:
             ~/.cargo/registry
             ~/.cargo/git
             futurenhs-platform/workspace-service/target
-          key: cargo-${{ hashFiles('futurenhs-platform/workspace-service/Cargo.lock') }}-a_
+          key: cargo-${{ hashFiles('futurenhs-platform/workspace-service/Cargo.lock') }}-1
```
alsuren, line 62: This feels like it has been done a bunch of times. Would it work to run `cargo clean --package workspace_service` like we do in `make prepare`?

frigus02: Not sure. I would like to get this in now to fix the master build. And then work on making sure this doesn't happen again.

frigus02: I hope this fixes it: https://github.com/FutureNHS/futurenhs-platform/pull/388
sqlx how cleans it's target/sqlx directory

cobwebfx (approved):

PR #388, opened 2020-10-19
Update sqlx to latest version from master

### LINKED ISSUE #371: Fix workspace-service caching issue
We should really find out what's going on here. Maybe we need to delete `target/sqlx` after build or something similar. But anyway, that's for another PR.

_Originally posted by @frigus02 in https://github.com/FutureNHS/futurenhs-platform/pull/368#discussion_r505527115_

frigus02:
Hopefully this fixes the sporadic cache issues we've been seeing with sqlx prepare.

Fixes #371

### DIFF
--- a/.github/workflows/workspace-service-branch.yaml
+++ b/.github/workflows/workspace-service-branch.yaml
@@ -48,10 +48,10 @@ jobs:
           path: |
             ~/.cargo/bin/sqlx
             ~/.cargo/bin/cargo-sqlx
-          key: sqlx-0.1.0-beta.1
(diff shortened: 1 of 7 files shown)


alsuren (approved):

frigus02:
I'm going to check if that really works. Will push a commit, which updates a query in a moment and see what it does.

frigus02:
This seems to fix `sqlx prepare --check` in CI. I also tested if we can remove `cargo clean` in our `make prepare` command, but that doesn't seem to work, yet. It still results in an empty `sqlx-data.json`.

frigus02:
This should only affect the build. I'd say we don't need QA on branch here.

PR #389, opened 2020-10-19
Update async-graphql to 2.0.7

frigus02:
Update async-graphql to 2.0.7. This replaces #383.

This update also required an update of:
- async-graphql-tide to 2.0.7
- tide to 0.14.0

I also noticed that files don't define a GraphQL entity and added that.

This update makes results of the introspection query deterministic. So I could remove the ugly custom sorting logic. 🎉

For some reason I couldn't run the tests on my machine. They immediately tried to connect to the COMPLETELY_BOGUS_DB_URL and failed with a timeout. I changed the function to use `connect_lazy` to fix the issue. Not sure why that didn't happen in CI.


frigus02:
We might be blocked on this issue: https://github.com/async-graphql/async-graphql/issues/315

It's also possible that I did something wrong in the update process.

frigus02:
The issue above got resolved super quickly. I now see another issue with the federation support: https://github.com/async-graphql/async-graphql/issues/316

workspace-service/src/graphql/files.rs
```diff
@@ -5,29 +5,30 @@ use chrono::{DateTime, Utc};
 use url::Url;
 use uuid::Uuid;
 
-#[SimpleObject(desc = "A file")]
+/// A file
+#[derive(SimpleObject)]
 pub struct File {
-    #[field(desc = "The id of the file")]
+    /// The id of the file
```
cobwebfx, line 11: Just wondering. What is the reason for 3 slashes?  I haven't seen this before.

cobwebfx (approved):

workspace-service/src/graphql/files.rs
frigus02: Three slashes is Rust's form of a doc comment (a comment, that makes it into generated documentation). As opposed to two slashes, which are just a normal code comment.

You can read all about this and more here: https://doc.rust-lang.org/rustdoc/ 🙂 

async-graphql now supports normal rust doc syntax for GraphQL objects. The custom `#[field(desc)]` macro has been removed.

PR #395, opened 2020-10-21
Change files table to make use of file versions

cobwebfx:
[AB#1798](https://dev.azure.com/futurenhs/75407963-f812-43e5-a734-1059bbc036a3/_workitems/edit/1798)

## Acceptance Criteria

- File table references file versions table
- All file related GraphQL functions use the new file versions table
- No breaking changes of the GraphQL API

## Pre-review checklist:

- [ ] Tests

## Testing information

We should upload a new file and make sure it:
- Still works
- Shows up in the file list in the folder
- The file page works
- The file download works

## Test:

- [x] Tested locally

### DIFF
--- a/workspace-service/graphql-schema.json
+++ b/workspace-service/graphql-schema.json
@@ -191,6 +191,22 @@
           {
             "args": [],
             "deprecationReason": null,
+            "description": "ID of the latest version of the file",
+            "isDeprecated": false,
+            "name": "latestVersion",
+            "type": {
+              "kind": "NON_NULL",
+              "name": null,
+              "ofType": {
+                "kind": "SCALAR",
+                "name": "ID",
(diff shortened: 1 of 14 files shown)


frigus02:
Do we need indexes or are foreign keys indexes by default?

> Since a DELETE of a row from the referenced table or an UPDATE of a referenced column will require a scan of the referencing table for rows matching the old value, it is often a good idea to index the referencing columns too. Because this is not always needed, and there are many choices available on how to index, declaration of a foreign key constraint does not automatically create an index on the referencing columns.
> -- https://www.postgresql.org/docs/current/ddl-constraints.html

So yes, we do. I added indexes for `file_versions(file)` and `file_versions(folder)`.

alsuren (approved):

PR #396, opened 2020-10-22
Better errors

StuartHarris:
This PR extracts the `message` property from the `ValidationErrors` and comma separates them for better display in the UI.
It also uses this in the tests, adding some more tests that exposed a bug in one of the validation regular expressions.

### DIFF
--- a/workspace-service/src/graphql/validation.rs
+++ b/workspace-service/src/graphql/validation.rs
@@ -1,0 +1,43 @@
+use std::fmt::Display;
+
+use validator::{ValidationErrors, ValidationErrorsKind};
+
+pub struct ValidationError(ValidationErrors);
+
+impl Display for ValidationError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        let errors = self.0.clone().into_errors();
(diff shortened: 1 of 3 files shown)


workspace-service/src/graphql/validation.rs
```diff
@@ -0,0 +1,41 @@
...
+use validator::{ValidationErrors, ValidationErrorsKind};
+
+const SEPARATOR: &str = ", ";
+
+pub struct ValidationError(ValidationErrors);
+
+impl Display for ValidationError {
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result {
+        let errors = self.0.clone().into_errors();
+        let errors = errors.iter().map(|(_, v)| v);
```
frigus02, line 11: Not sure if we have to clone here since we clone the individual errors later already. The `HashMap` also has a shorthand to get a list of values. So we can replace these 2 lines with:
```suggestion
        let errors = self.0.errors().values();
```

frigus02 (approved):

workspace-service/src/graphql/validation.rs
StuartHarris: nice! Thanks

PR #397, opened 2020-10-22
Design sweep

matt-paul:
[Ticket]()

[Zeplin]()

## Pre-review checklist:

- [ ] Tests

- [ ] Documentation

- [ ] Feature flag

## Testing information

- Test plan?

## Test:

- [ ] Tested locally

- Platforms/Browsers:

  - [ ] Internet Explorer 11

  - [ ] Edge

- Accessibility:

  - [ ] Screen Reader

  - [ ] Keyboard Navigation

  - [ ] Magnification

  - [ ] Contrast

  - [ ] Text size 200%

  - [ ] Landscape (Mobile)


frontend/components/H2/H2.tsx
```diff
@@ -0,0 +1,8 @@
+import React from "react";
+
+interface Props {
+  title: string;
+}
+const H2 = ({ title }: Props) => <h2>{title}</h2>;
```
frigus02, line 6: Since this component has no styles, I'm inclined to remove it and use `h2` directly. 🤔

frontend/components/Workspaces/WorkspaceDirectoryItem.tsx
```diff
@@ -53,13 +44,13 @@ interface Props {
...
+        alt=""
+      />
       <Link href="/workspaces/[id]" as={`/workspaces/${id}`} passHref>
         <StyledLink>
-          <img
-            src={require("../../public/Placeholder_Workspace_Image.svg")}
-            alt=""
-          />
-          <h3>{title}</h3>
+          <h4>{title}</h4>
```
frigus02, line 53: I wonder if those should actually be headings. It doesn't make sense to me semantically. They don't indicate a new section on the page. If this would show the workspace description, then maybe it would make sense to me. But at the moment, I think this should just be a div/span.

frigus02 (commented):
Code looks nice. Just 2 minor comments. I have not tested it in the browser, yet. But I assume you did that with Kim a lot 😄

frontend/components/H2/H2.tsx
matt-paul: Yeah, the reasoning there was that it means they will stay aligned in future and stop drift across pages

matt-paul: no biggie though, happy to change if preferred

frigus02: Hm yeah. I'm generally for keeping as simple as we can. And we don't need it at the moment and don't know for sure if we will need it in the future.

But I do see your point. Styles already drifted apart a little. I'm okay to keep it as well, if we think that keeps the styles more consistent.

frontend/components/Workspaces/WorkspaceDirectoryItem.tsx
matt-paul: yeah i agree actually, will change that

frigus02 (approved):

declanslevin:
The new meatball component is missing hover/focus/active states

frigus02:
Good catch. We're going to fix that now

PR #398, opened 2020-10-22
User email_address 

trcywu:
subtask of [AB#2072](https://dev.azure.com/futurenhs/75407963-f812-43e5-a734-1059bbc036a3/_workitems/edit/2072) 

[Zeplin]()

## Acceptance Criteria

- [x] inserted email_address column to users table

## Pre-review checklist:

- [ ] Tests

- [ ] ~Documentation~

- [ ] ~Analytics (user analytics, e.g. Google Analytics)~

- [ ] ~Observability (metrics/tracing)~

- [ ] ~Feature flag~

- [ ]~ Data columns adhere to the [Dublin Core Metatdata schema](https://www.dublincore.org/specifications/dublin-core/dcmi-terms/). Further information is available [here](https://www.gov.uk/government/publications/recommended-open-standards-for-government/using-metadata-to-describe-data-shared-within-government)~

## Testing information

- Test plan
- Test Login in production. 
## Test:

- [ ] Tested locally


workspace-service/migrations/20201022135333_groups.sql
```diff
@@ -0,0 +1,18 @@
+CREATE TABLE IF NOT EXISTS groups (
```
StuartHarris, line 1: `groups` and `user_groups` are now in the other PR, so should they be removed from here? (and `group_workspace_grants` put into a new migration file)

alsuren:
I'm going to try getting this rebased and finished off. I think it needs to be joined up all the way to the frontend, so that the email address from azure is propagated down on user creation.

frontend/server.js
```diff
@@ -118,6 +118,7 @@ async function main() {
               const response = await getOrCreateUser({
                 authId: profile.sub,
                 name: profile.displayName,
+                emailAddress: profile.emails,
```
frigus02, line 121: I think `emails` is an array, so this should be:

```suggestion
                emailAddress: profile.emails[0],
```

frontend/server.js
```diff
@@ -118,6 +118,7 @@ async function main() {
...
               });
               const {
```
frigus02, line 123: I wonder if we should use the email address returned from here for the user profile, similar to how we do this in the Noop strategy.

frontend/lib/server/graphql.js
```diff
@@ -3,22 +3,28 @@ const { createClient } = require("urql");
 
 const { requireEnv } = require("./requireEnv");
 
-const getOrCreateUser = async ({ authId, name }) => {
+//todo: add email address in
```
frigus02, line 6: Looks like this todo is done.

workspace-service/migrations/20201022144035_user-email.sql
```diff
@@ -0,0 +1,7 @@
+-- Add migration script here
+ALTER TABLE users
+  ADD COLUMN email_address TEXT NOT NULL DEFAULT ('testuser@example.com');
```
frigus02, line 3: I'm not sure about this default value. The users we have in production are not test users. I think I would either put in some email address indicating that the address does not exist (maybe `unknown@localhost`) or make the column nullable.

workspace-service/migrations/20201022150109_insert-users.sql
```diff
@@ -0,0 +1,59 @@
+INSERT into users (
```
frigus02, line 1: I like the colorful names. 😄 But why do we need these test users?

alsuren: Might need to check with @StuartHarris, but I think that this is a reasonable idea.

On the other PR, we found it really fiddly to develop/test the workspace admins/members migration because `make reset-local` causes you to end up with no workspaces (we ended up making a temporary migration and not checking it in). Where it is easy, I am in favour of migrations to insert example records in the DB. Having the default admin user and a handful of test users makes sense to me.

frontend/server.js
alsuren: Ah. That might explain why the login flow on https://fnhs-dev-david.westeurope.cloudapp.azure.com/auth/login fails to complete after I deployed this branch.

workspace-service/migrations/20201022150109_insert-users.sql
frigus02: Yeah, I agree. I can see this being useful if we extend the Noop login strategy in the frontend, so that we can choose one of those users.

Although we could say a similar thing about workspaces, folders and files. For those we have shell scripts. Maybe we should have shell scripts for the users as well? I thought we could possibly use those scripts for migration later on. But that might not be realistic.

frontend/server.js
trcywu: what do you mean?

frigus02: Sorry, GitHub didn't let me add a comment at the relevant place. We're returning the session profile a few lines down:

```js
              const {
                id,
                name,
                authId,
                isPlatformAdmin,
              } = response.getOrCreateUser;
              done(null, {
                id,
                authId,
                name,
                isPlatformAdmin,
                emails: profile.emails,
              });
```

And we're setting`emails` to `profile.emails`. I wonder if we should rather extract `emailAddress` from the GraphQL getOrCreateUser response and use that instead: `emails: response.getOrCreateUser.emailAddress`.

With this we would use an email address that has been validated by our backend.

trcywu: ok right. is it the same if i just deconstruct emailAddress from `response.getOrCreateUser` and pass it through? 
```   async (profile, done) => {
            try {
              const response = await getOrCreateUser({
                authId: profile.sub,
                name: profile.displayName,
                emailAddress: profile.emails[0],
              });
              const {
                id,
                name,
                authId,
                isPlatformAdmin,
                emailAddress,
              } = response.getOrCreateUser;
              done(null, {
                id,
                authId,
                name,
                isPlatformAdmin,
                emails: emailAddress,
              });
            } catch (err) {
              done(err);
            }
          }```

frigus02: Yeah, exactly. I just wrote it out to make it clearer what I mean. We should go with your code

frigus02 (approved):
I guess this is okay. I'm still on the fence with the test users. We have 2 different ways now of creating test data:

- SQL migrations
- Shell scripts

Shell (or any other external) script seems cleaner to me and they can handle all cases. For example it's impossible to insert files with DB migrations. But I get that migrations are much more convient because they happen automatically. I don't want to block that. We can always change it later.

alsuren (approved):

PR #400, opened 2020-10-23
Group tables migration

trcywu:
[Ticket]()

[Zeplin]()

## Acceptance Criteria

- [x] created migrations for 'group', 'user_groups'
- [x] add foreign key for members & admins but allow them to be null
- [x] update create workspace query to also create a group
- [x] create migration to fill in groups for all workspaces
- [x] add not null constraint back in

## Pre-review checklist:

- [x] Tests

- [ ] ~Documentation~

- [ ] ~Analytics (user analytics, e.g. Google Analytics)~

- [ ] ~Observability (metrics/tracing)~

- [ ] ~Feature flag~

- [ ] ~Data columns adhere to the [Dublin Core Metatdata schema](https://www.dublincore.org/specifications/dublin-core/dcmi-terms/). Further information is available [here](https://www.gov.uk/government/publications/recommended-open-standards-for-government/using-metadata-to-describe-data-shared-within-government)~

## Testing information

- Test plan

on your branch run: 
- `cd workspace-service/ && make reset-local run-local` 
-  then run `./workspace-service/scripts/create-default-test-folders.sh local` in another tab
- go to tableplus to check that the groups table is created. And the admin and members columns are linked correctly with the correct names from workspaces table

## Test:

- [x] Tested locally

### DIFF
--- a/workspace-service/migrations/20201022135333_groups.sql
+++ b/workspace-service/migrations/20201022135333_groups.sql
@@ -12,6 +12,7 @@ CREATE TABLE IF NOT EXISTS user_groups (
 
 ALTER TABLE workspaces ADD COLUMN members uuid REFERENCES groups;
 
+
 ALTER TABLE workspaces ADD COLUMN admins uuid REFERENCES groups;
 
 WITH groups_to_insert AS (
@@ -30,3 +31,24 @@ SELECT
 	title
 FROM
 	groups_to_insert;
+
+  WITH groups_to_insert AS (
+	UPDATE
+		workspaces
+	SET
+		members = uuid_generate_v4 ()
+	WHERE
+		members ISNULL
+	RETURNING
+		members AS id,
+		title
+) INSERT INTO GROUPS (id, title)
+SELECT
+	id,
+	title
+FROM
+	groups_to_insert;
+
+--the last 2 alter would usually be done in a separate migration
+ALTER TABLE workspaces ALTER COLUMN members SET NOT NULL;
+;
--- a/workspace-service/migrations/20201022135333_groups.sql
+++ b/workspace-service/migrations/20201022135333_groups.sql
@@ -13,3 +13,20 @@ CREATE TABLE IF NOT EXISTS user_groups (
 ALTER TABLE workspaces ADD COLUMN members uuid REFERENCES groups;
 
 ALTER TABLE workspaces ADD COLUMN admins uuid REFERENCES groups;
+
+WITH groups_to_insert AS (
+	UPDATE
+		workspaces
+	SET
+		admins = uuid_generate_v4 ()
+	WHERE
+		admins ISNULL
+	RETURNING
+		admins AS id,
+		title
+) INSERT INTO GROUPS (id, title)
+SELECT
+	id,
+	title
+FROM
+	groups_to_insert;
--- a/workspace-service/migrations/20201022135333_groups.sql
+++ b/workspace-service/migrations/20201022135333_groups.sql
@@ -1,6 +1,6 @@
 CREATE TABLE IF NOT EXISTS groups (
   id uuid DEFAULT uuid_generate_v4 (),
-  name TEXT NOT NULL,
+  title TEXT NOT NULL,
   PRIMARY KEY (id)
 );
 
--- a/workspace-service/migrations/20201022135333_groups.sql
+++ b/workspace-service/migrations/20201022135333_groups.sql
@@ -4,15 +4,12 @@ CREATE TABLE IF NOT EXISTS groups (
   PRIMARY KEY (id)
 );
 
-CREATE TABLE IF NOT EXISTS group_workspace_grants (
-  is_admin BOOLEAN NOT NULL,
-  group_id uuid NOT NULL REFERENCES groups,
-  workspace_id uuid NOT NULL REFERENCES workspaces,
-  CONSTRAINT group_workspace_grant PRIMARY KEY (is_admin, group_id, workspace_id)
-);
-
 CREATE TABLE IF NOT EXISTS user_groups (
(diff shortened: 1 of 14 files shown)


workspace-service/migrations/20201022135333_groups.sql
```diff
@@ -0,0 +1,18 @@
...
+  id uuid DEFAULT uuid_generate_v4 (),
+  name TEXT NOT NULL,
+  PRIMARY KEY (id)
+);
+
+CREATE TABLE IF NOT EXISTS group_workspace_grants (
+  is_admin BOOLEAN NOT NULL,
+  group_id uuid NOT NULL REFERENCES groups,
+  workspace_id uuid NOT NULL REFERENCES workspaces,
+  CONSTRAINT group_workspace_grant PRIMARY KEY (is_admin, group_id, workspace_id)
```
frigus02, line 11: Should `is_admin` be part of the primary key? This would mean a group+workspace could be in here twice: once with is_admin:false and once with is_admin:true.

frigus02: Ah right. Does this mean we have 2 groups per workspace?
- One admin group
- One member group (is_admin = false)

alsuren: ... now that I think of it, we were also thinking of making is_admin into an enum (admin|allmembers) to make things clearer

alsuren: the constraint is "there can only be one admin group per workspace". Might be there is a better way to do this.

alsuren: Just talking to James, and we think that the constraint should be more like:

there must be exactly one admin group per workspace
there must be exactly one all users group per workspace

These should just be added as fields on the workspace.

frigus02: Yeah, that makes sense. An enum will make this easier to understand I think.

frigus02: > These should just be added as fields on the workspace.

That makes sense to me as well

workspace-service/migrations/20201022135333_groups.sql
```diff
@@ -30,3 +31,24 @@ SELECT
...
+) INSERT INTO GROUPS (id, title)
+SELECT
+	id,
+	title
+FROM
+	groups_to_insert;
+
+--the last 2 alter would usually be done in a separate migration
+ALTER TABLE workspaces ALTER COLUMN members SET NOT NULL;
+;
```
alsuren, line 54: ah. The second ALTER TABLE went missing. Maybe that explains the errors about Options

workspace-service/src/db/groups.rs
```diff
@@ -0,0 +1,84 @@
...
+impl Group {
+    pub async fn create(title: &str, pool: &PgPool) -> Result<Group> {
+        let group = sqlx::query_file_as!(Group, "sql/groups/create.sql", title)
+            .fetch_one(pool)
+            .await?;
+
+        Ok(group)
+    }
+    #[allow(dead_code)]
+    pub async fn find_by_id(id: Uuid, pool: &PgPool) -> Result<Group> {
```
alsuren, line 23: note to reviewer: these are dead code because we've not written the graphql bit yet. It's mostly copy-pasted CRUD boilerplate, so didn't take too long to write. We can do a dead code sweep once we have the graphql query code written.

workspace-service/src/db/workspaces.rs
```diff
@@ -9,15 +11,26 @@ pub struct Workspace {
...
+        let workspace = sqlx::query_file_as!(
+            Workspace,
+            "sql/workspaces/create.sql",
+            title,
+            description,
+            admins.id,
+            members.id
+        )
+        .fetch_one(pool)
+        .await?;
```
frigus02, line 33: I feel like we should do this in a database transaction. Example code is in files:

https://github.com/FutureNHS/futurenhs-platform/blob/f81af90788ce04aef64631d3dccdba3a1d26cbfd/workspace-service/src/db/files.rs#L122-L140

And for the executor types:

https://github.com/FutureNHS/futurenhs-platform/blob/f81af90788ce04aef64631d3dccdba3a1d26cbfd/workspace-service/src/db/files.rs#L22-L24

workspace-service/src/db/groups.rs
frigus02: I wonder if we need `find_by_id`, `update` and `delete` at all.

Doing a sweep at the end to see which ones we actually end up using sounds good to me. Given the current columns, we might be able to remove them.

frigus02 (commented):
Looks good. I think a DB transaction would be good. Otherwise 👍

workspace-service/src/db/groups.rs
alsuren: in the end I couldn't be bothered to do the E: Executor dance for them all so we dropped them on the floor.

workspace-service/src/db/workspaces.rs
```diff
@@ -9,15 +11,29 @@ pub struct Workspace {
...
 impl Workspace {
     pub async fn create(title: &str, description: &str, pool: &PgPool) -> Result<Workspace> {
-        let workspace =
-            sqlx::query_file_as!(Workspace, "sql/workspaces/create.sql", title, description)
-                .fetch_one(pool)
-                .await?;
+        let mut tx = pool.begin().await?;
+
+        let admins = Group::create(&format!("{} Admins", title), &mut tx).await?;
+        let members = Group::create(&format!("{} Members", title), &mut tx).await?;
```
frigus02, line 24: Nice. Adding "Admins", "Members" to the title is a good idea 👍

frigus02 (approved):

PR #401, opened 2020-10-23
Upload form for file version

frigus02:
[AB#1798](https://dev.azure.com/futurenhs/75407963-f812-43e5-a734-1059bbc036a3/_workitems/edit/1798)

## Acceptance Criteria

Refactor upload form so it can be used for uploading files and uploading a new file version

- [x] Component for remaining characters
- [x] input[type=file] --> remove "multiple" and change "hint"
- [x] title input --> change label and prepopulate with original file title instead of selected file name
- [x] description input --> change label and prepopulat with original file description
- [x] onSubmit --> call createFileVersion instead of createFile
- [x] onSubmit success --> redirect to file page instead of folder
- [x] update cache

## Testing information

- Fileupload UI should not have changed but has been refactored, so probably worth a look.

### DIFF
--- a/frontend/containers/UploadFileForm/UpdateFileForm.tsx
+++ b/frontend/containers/UploadFileForm/UpdateFileForm.tsx
@@ -36,7 +36,7 @@ interface Props {
   workspaceId: string;
 }
 
-const SingleUploadFileForm: FC<Props> = ({
+const UpdateFileForm: FC<Props> = ({
   fileDescription,
   fileId,
   fileTitle,
@@ -172,4 +172,4 @@ const SingleUploadFileForm: FC<Props> = ({
   );
 };
 
-export default SingleUploadFileForm;
+export default UpdateFileForm;
--- a/frontend/containers/UploadFileForm/UpdateFileForm.tsx
+++ b/frontend/containers/UploadFileForm/UpdateFileForm.tsx
@@ -90,7 +90,9 @@ const UpdateFileForm: FC<Props> = ({
             throw new Error(`Failed to save file: ${file.error?.message}`);
           }
 
-          router.push(`/workspaces/${workspaceId}/folders/${folderId}`);
+          router.push(
+            `/workspaces/${workspaceId}/folders/${folderId}/files/${fileId}`
+          );
         } catch (error) {
           setError(`fileData.title`, {
             type: "server",
--- a/frontend/containers/UploadFileForm/UpdateFileForm.tsx
+++ b/frontend/containers/UploadFileForm/UpdateFileForm.tsx
@@ -94,7 +94,7 @@ const UpdateFileForm: FC<Props> = ({
             `/workspaces/${workspaceId}/folders/${folderId}/files/${fileId}`
           );
         } catch (error) {
-          setError(`fileData.title`, {
+          setError("fileData.title", {
             type: "server",
             message: error.toString(),
           });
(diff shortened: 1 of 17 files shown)


frontend/containers/UploadFileForm/UpdateFileForm.tsx
```diff
@@ -0,0 +1,177 @@
...
+
+          if (file.error) {
+            throw new Error(`Failed to save file: ${file.error?.message}`);
+          }
+
+          router.push(
+            `/workspaces/${workspaceId}/folders/${folderId}/files/${fileId}`
+          );
+        } catch (error) {
+          setError(`fileData.title`, {
```
alsuren, line 97:
```suggestion
          setError("fileData.title", {
```
(backticks seem unnecessary. I've not read what setError is doing, so might be that this should be different in some other way)

frontend/lib/useMaxLengthHelper.ts
```diff
@@ -0,0 +1,35 @@
+import { useState } from "react";
+
+export const useMaxLengthHelper = (fieldTitle: string, maxLength: number) => {
```
alsuren, line 3: can we have some return types here? (I feel like eslint should be able to catch implicit `any` return types. Do we have it misconfigured?)

frigus02: This isn't `any`. It's an inferred return type based on the code in the function. I'm personally fine with this here, because I don't really care about the type. I just care that it's used correctly, which is verified by TypeScript. But if you prefer to write it out explicitly here, we can do that.

alsuren: ah. Typescript continues to amaze me with what it's happy to infer the types of.

frigus02: Yeah.  We do have eslint for implicit `any`. So that should be cought automatically if we ever do that accidentally.

alsuren (approved):
I *think* it looks good, but I only have a surface knowledge of this area of the codebase. I've asked Tracy to take a look as well.

frontend/containers/UploadFileForm/UpdateFileForm.tsx
```diff
@@ -0,0 +1,177 @@
...
+    } catch (error) {
+      setError("files", {
+        type: "server",
+        message: error.toString(),
+      });
+    }
+  };
+  return (
+    <Form id="filesUploadForm" onSubmit={handleSubmit(onSubmit)}>
+      <StyledInput
```
trcywu, line 112: Is there an opportunity to break down and extract components from the `UploadFileForm` and be reused here? Feels like both are `UpdateFileForm` and `UploadFileForm` are very similar?

frigus02: They are, yes. We extracted all the UI components (like `StyledInput`). But we can't easily extract the entire component because they all differ in at least 1 prop:

- The file input has a different `multiple`, `hint` and `label`
- The inputs for title and description have different `label` and `defaultValue`

It feels to me like this is the best we can do without making those 2 upload forms too dependent on each other. I don't want to end up with a component with too many props. I think that can be quite painful. But at least the CSS is shared 🙂

trcywu: ok cool. Seems like it's been attempted and thought through then. Looking good, especially love the helper!

trcywu (approved):

frontend/containers/UploadFileForm/UpdateFileForm.tsx
frigus02: Good to ask, though. I would definitely extract more common components, I just couldn't find any 😄

PR #417, opened 2020-10-26
Local Azure Blob Storage emulator

frigus02:
Our workspace service can work with a local PostgreSQL database. But it still requires an Azure Blob Storage account with 2 containers. This is a bit of a hurdle when trying to get up and running as a new (and possibly external) contributor.

This PR adds a way of using [Azurite](https://github.com/Azure/Azurite), the new cross platform Azure Storage emulator, locally.

It works but I had to add quite a few workarounds to make it compatible with the Azure SDK for Rust. I will create issues in both the Azure SDK for Rust and Azurite to understand how to set this up correctly. Those issues are:

- [ ] The Azure SDK for Rust supports the emulator but has a special way to construct a client. I think it would be super handy if it could understand the Azure Storage connection string format. Then we could supply this in an environment variable and be good to go.

- [x] The SAS generation in the Azure SDK does not work for emulator URLs. Since the emulator runs on localhost the account name is part of the path, not a subdomain.

  Solved by using `KeyClient` instead of `BlobSASBuilder`.

- [ ] Azure does not return a "x-ms-copy-status" response header for the CopyBlobByUrl API but the Azure SDK for Rust requires it. I'm not sure which one if correct. https://github.com/Azure/Azurite/issues/603

Anyway, this works. Just `make run-local` and you're good to go 🎉 

I think the code needs a bit of cleanup but we could possibly get this merged with the workarounds in place.


frigus02:
> The SAS generation in the Azure SDK does not work for emulator URLs. Since the emulator runs on localhost the account name is part of the path, not a subdomain.

At the moment we're using the [`BlobSASBuilder`](https://docs.rs/azure_sdk_storage_core/0.44.4/azure_sdk_storage_core/prelude/struct.BlobSASBuilder.html), which does not support the emulator. I wonder if we can use [`KeyClient::shared_access_signature`](https://docs.rs/azure_sdk_storage_core/0.44.4/azure_sdk_storage_core/key_client/struct.KeyClient.html#impl-ClientSharedAccessSignature) instead, which does support the emulator.

workspace-service/Makefile
```diff
@@ -14,14 +16,14 @@ run: ## Fetch DB URL and run locally
...
 
 .PHONY: run-local
-run-local: migrate-local ## Fetch DB URL and run locally. Does not have file support yet.
-	UPLOAD_CONTAINER_URL=https://obviously.fake \
-	FILES_CONTAINER_URL=https://obviously.fake \
-	FILE_STORAGE_ACCESS_KEY=also-fake \
+run-local: migrate-local ## Run locally against Postgres and Azure running in Docker
 	DATABASE_URL=$(localdb) \
+	UPLOAD_CONTAINER_URL=http://127.0.0.1:10000/devstoreaccount1/upload \
+	FILES_CONTAINER_URL=http://127.0.0.1:10000/devstoreaccount1/files \
```
alsuren, line 25: at my previous company we got people to entries to /etc/hosts for this kind of thing. Should we change this to a hostname that parses correctly and then add a make target to assert that /etc/hosts is set up correctly?

frigus02: I thought about this. But we still need the account name in the path because Azurite expects that. Let's say I put this in my `/etc/hosts` file:

```
127.0.0.1 devstoreaccount1.blob.local
```

Then:

- this works: http://devstoreaccount1.blob.local:10000/devstoreaccount1/files/3c8bb5bf-55d8-4200-8b1b-c11b1e27f703
- this doesn't work: http://devstoreaccount1.blob.local:10000/files/3c8bb5bf-55d8-4200-8b1b-c11b1e27f703

So we still need logic to parse the URL differently. But we don't gain much, I think. The connection string supports specifying URLs. We can make a PR to the Azure SDK for Rust to use those. Then the only problem is the missing copy header. I will create an issue for this now.

workspace-service/src/azure/blob.rs
```diff
@@ -73,24 +74,33 @@ pub async fn copy_blob_from_url(url: &Url, azure_config: &super::Config) -> Resu
...
             .with_is_synchronous(true)
             .finalize(),
     )
-    .await?;
-
-    match response.copy_status {
+    .await;
+    let copy_status = match result {
+        Ok(response) => Ok(response.copy_status),
+        Err(AzureError::HeaderNotFound(header))
```
alsuren, line 90: Can we have a comment linking to the upstream bugs for this? This will make it easier for anyone to subscribe to them and clean this up once they're fixed upstream.

workspace-service/src/azure/mod.rs
```diff
@@ -1,24 +1,62 @@
...
+            account: upload.account,
+            upload_container: upload.container,
             upload_container_url,
+            files_container: files.container,
             files_container_url,
+        })
+    }
+
+    pub fn is_emulator(&self) -> bool {
+        self.access_key == DEFAULT_EMULATOR_ACCESS_KEY
```
alsuren, line 49: - [ ] is_local_emulator()? (we seem to be using `-local` in the make targets, so it will be a smaller mental leap if this has `local` in the name somewhere.

- [ ] either add a comment linking this constant to the one in the makefile or go based on the localhostness of the container url?

frigus02: >  is_local_emulator()? (we seem to be using -local in the make targets, so it will be a smaller mental leap if this has local in the name somewhere.

👍 

>  either add a comment linking this constant to the one in the makefile or go based on the localhostness of the container url?

That constant actually comes from Azurite. Probably a good idea to link to that documentation: https://github.com/Azure/Azurite#default-storage-account

workspace-service/src/azure/sas.rs
```diff
@@ -17,31 +17,55 @@ fn create_upload_sas_impl(config: &Config, name: &Uuid, now: DateTime<Utc>) -> R
...
-    Ok(sas)
+    let token = config
+        .client()
+        .shared_access_signature()
+        .with_start(start)
+        .with_expiry(end)
+        .with_permissions(SasPermissions::Write)
+        .with_resource(SasResource::Blob)
+        .with_resource_type(SasResourceType::Object)
+        .with_protocol(if config.is_emulator() {
```
alsuren, line 28: `.with_protocol(config.protocol())` ? Might be a step too far. I'm not sure.

frigus02: Oh yeah. I think that's nicer 👍

alsuren:
Tested locally and it seems to work. Are we getting to the point that we could run `workspace-service/scripts/create-default-test-folders.sh local` (or even a test migration that uploads real files) in CI?

frigus02:
Yeah, after this we should be able to upload files in CI. We should update the shell scripts to actually upload some dummy files. Sounds like a good follow-up.

workspace-service/src/azure/mod.rs
frigus02: I added a link to the Azurite docs both for the `DEFAULT_EMULATOR_ACCESS_KEY` and in the Makefile.

alsuren (approved):

frigus02:
I'm going to merge this. And make sure it all still works in master 🤞

PR #420, opened 2020-10-27
FileCreated and FileUpdated events

frigus02:
[Ticket](https://dev.azure.com/futurenhs/FutureNHS/_boards/board/t/FutureNHS%20Team/Stories/?workitem=1798)

This PR contains the final task in the ticket - adding events for 'file created' and 'file updated'.

## Pre-review checklist:

- [x] Tests

NOTE: No QA is needed.

### DIFF
--- a/workspace-service/src/graphql/files.rs
+++ b/workspace-service/src/graphql/files.rs
@@ -1,7 +1,9 @@
 use super::{azure, db, validation, RequestingUser};
 use async_graphql::{Context, FieldResult, InputObject, Object, SimpleObject, ID};
-use fnhs_event_models::{Event, EventClient, EventPublisher as _, FileCreatedData};
 use chrono::{DateTime, Utc};
+use fnhs_event_models::{
+    Event, EventClient, EventPublisher as _, FileCreatedData, FileUpdatedData,
+};
 use lazy_static::lazy_static;
 use mime_db::extensions2;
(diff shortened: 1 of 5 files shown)


workspace-service/src/graphql/files.rs
```diff
@@ -271,18 +286,35 @@ async fn create_file(
...
+            file.id.clone(),
+            FileCreatedData {
+                file_id: file.id.to_string(),
+                created_at: file.created_at.to_rfc3339(),
+                file_description: file.description.clone(),
+                file_title: file.title.clone(),
+                file_type: file.file_type.clone(),
+                folder_id: folder_id.to_string(),
+                user_id: user.id.to_string(),
+                workspace_id: folder.workspace.to_string(),
```
alsuren, line 303: workspace_id is not included in the tracking requirements on the ticket, and it seems a bit of a rigmarole to get it. Are we sure we actually need it?

On the other hand, file upload operations are rare and slow anyway, so an extra db query isn't going to hurt the user experience at all.

alsuren (approved):
🏃 🛤️

workspace-service/src/graphql/files.rs
matt-paul: Yeah, we went through the requirements again at the start of the PR with Jamie and he requested it be there.

PR #422, opened 2020-10-27
remove snapshot tests of pages/*

alsuren:
When mob-reviewing, we came to the conclusion that deep-snapshot-testing
pages causes a lot of pointless snapshot churn.

The functionality of the pages is tested in cypress anyway, so these
'does not explode' tests might not be super-valuable, but I argued
to keep them because jest has a quicker feedback loop than cypress

Once we are at a stage of the project where we have users and want
to be really careful about introducing regressions to key pages,
then we can think about turning them back into snapshot tests.


frigus02 (approved):
I like this ❤️ 

Makes we wonder where to draw the line to Cypress tests.

matt-paul:
🙌

matt-paul (approved):

PR #423, opened 2020-10-27
make fmt check; document make graphql-schema.json; explicit postgres config in docker-compose

alsuren:
Just things that I noticed while pairing

Also make postgres config explict in docker-compose.yaml

I couldn't find a gif of someone making something with a file. Sorry.

### DIFF
--- a/workspace-service/Makefile
+++ b/workspace-service/Makefile
@@ -24,6 +24,14 @@ run-local: migrate-local ## Fetch DB URL and run locally. Does not have file sup
 	DATABASE_URL=$(localdb) \
 		$(MAKE)
 
+.PHONY: check
+check: ## Run `cargo check` (because David keeps saying "make check" when he means "cargo check").
+	cargo check
+
+.PHONY: fmt
+check: ## Run `cargo fmt` (convenience so that you can write `make fmt check test` in one command)
+	cargo fmt
+
 .PHONY: test
 test: ## Run tests [TEST=test_name (optional)]
 	cargo test $$TEST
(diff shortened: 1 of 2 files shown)


workspace-service/Makefile
```diff
@@ -24,6 +24,14 @@ run-local: migrate-local ## Fetch DB URL and run locally. Does not have file sup
 	DATABASE_URL=$(localdb) \
 		$(MAKE)
 
+.PHONY: check
+check: ## Run `cargo check` (because David keeps saying "make check" when he means "cargo check").
+	cargo check
+
+.PHONY: fmt
+check: ## Run `cargo fmt` (convenience so that you can write `make fmt check test` in one command)
```
frigus02, line 32:
```suggestion
fmt: ## Run `cargo fmt` (convenience so that you can write `make fmt check test` in one command)
```

workspace-service/Makefile
```diff
@@ -24,6 +24,14 @@ run-local: migrate-local ## Fetch DB URL and run locally. Does not have file sup
 	DATABASE_URL=$(localdb) \
 		$(MAKE)
 
+.PHONY: check
+check: ## Run `cargo check` (because David keeps saying "make check" when he means "cargo check").
+	cargo check
```
frigus02, line 29: Should we use `clippy` here? That reminds me that I wanted to look up how much more `clippy` does compared to `check` and if it does more at all.

alsuren: `check` is the fastest way to typecheck your code.

In my mind, clippy is a set of linting rules that you can run after everything typechecks, to suggest best-practices. It hadn't occurred to me to run clippy instead of check. I don't know what it does if your code doesn't even compile.

frigus02: Oh right. Well, let's keep check here. It's probably useful.

frigus02 (approved):

PR #431, opened 2020-10-29
❓ use db::Workspace everywhere, and don't have a graphql::Workspace struct

alsuren:
Crazy idea: Now that Workspace is no longer a SimpleObject, we can probably get away without a graphql::Workspace struct entirely. The methods on `#[Object] impl Workspace` replace the need for a `impl From<db::Workspace> for Workspace {` and the GraphQL schema stays the same.

If we're doing this then we should *probably* apply the same treatment to all of our models.

Upsides include:

* No need to keep track of which `Workspace` struct we mean.
* No mostly-passthrough .into() implementation.
* Easier and more natural to extend our objects with resolvers that optionally fetch from the DB at a later date.

Downsides include:

* We have an `impl Struct` that doesn't live in the same place as `struct Struct {}`
  * Splitting impl blocks up into different modules is actually surprisingly common in rust-land, so this might just be fine.
* The `#[Object]` attribute macro is super-magical, and the methods you write end up with a completely different signature from what you expect (Workspace already has this problem, but we would be extending it to all of our models).

kill it with fire?

### DIFF
--- a/workspace-service/Makefile
+++ b/workspace-service/Makefile
@@ -34,8 +34,8 @@ check: ## Run `cargo check` (because David keeps saying "make check" when he mea
 	cargo check
 
 .PHONY: clippy
-clippy: ## Run `cargo clippy`, the Rust linter.
-	cargo clippy
+clippy: ## Run `cargo clippy`, the Rust linter. --release is to get around some interactions with cargo check.
+	cargo clippy --release -- -D warnings
 
 .PHONY: fmt
 fmt: ## Run `cargo fmt` (convenience so that you can write `make fmt check test clippy` in one command)
--- a/workspace-service/src/graphql/files.rs
+++ b/workspace-service/src/graphql/files.rs
@@ -1,5 +1,5 @@
-use super::{azure, db, validation, RequestingUser};
-use async_graphql::{Context, FieldResult, InputObject, Object, SimpleObject, ID};
+use super::{azure, db, db::FileWithVersion, validation, RequestingUser};
+use async_graphql::{Context, FieldResult, InputObject, Object, ID};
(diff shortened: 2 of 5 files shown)


frigus02:
I'm not sure how I feel about this. I think my main concern is that lazyness and copy-pasting a pattern like this could lead to:

- Accidental changes of the GraphQL schema, when the DB schema changes. That's probably not a massive deal. We could create tests for that if it becomes a problem.
- Not optimal GraphQL or DB schema/types, because we're trying to make both sides work with one struct (e.g. Uuid vs ID types).

However I'm generally all for making things simpler if possible. And that does seem simpler to me, at least for the `Workspace` now.

StuartHarris:
I think it is OK for `#[Object]` implementations where the underlying struct seems to have no bearing on the GraphQL schema and, I guess, is really only used to carry data to the resolver methods?
Obvs wouldn't be good for `#[derive(SimpleObject)]`.

alsuren:
Pushed a commit to apply the Object treatment to File. It's a bit odd because it converts from FileWithVersion, but it would allow you to consider adding a `versions` resolver, so you can say:
```
query file(id: "abc") {
    versions {
        created_at
        created_by
    }
}
```
or something later.

I haven't dug very deeply into how the db::File* structs fit together, so I've just done the same mechanical translation as the other commits, and used `#[Object(name = "File")]` to do the rename. Might be that there is a better way to do this in a separate PR.

alsuren:
After the weekend, I'm still on the fence. Do we merge it and stop using SimpleObject entirely, or close it and only convert things to Object when they need complicated resolvers?

I think that the answer is close it and convert things as we need to. The conversion isn't hard.

PR #433, opened 2020-10-29
Enable demo data file upload script

frigus02:
Now that we have file uploads and the local Azure Storage emulator working, we can enable the demo data script, that uploads files. 🎉


workspace-service/scripts/_mime_type.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+set -euo pipefail
+
+_get_mime_type() {
+	FILE_NAME="$1"
+	if [[ "$FILE_NAME" == *.csv ]]; then
+		echo "text/csv"
```
frigus02, line 8: `file` only works based on magic numbers in the file, not based on the file extension. This means it reports `text/plain` for CSV. I couldn't find a generally available tool that does that. But this manual check seems okay to me.

StuartHarris:
@frigus02 Thanks for fixing my bug. I added `local` env to the usage examples, as I was a bit confused when I was trying it out. Looks great though

frigus02:
Yeah, good catch about the usage example.

Looks like `shfmt` isn't happy with some of the formatting now. I think it might prefer tabs over spaces.

workspace-service/scripts/create-default-test-folders.sh
```diff
@@ -1,31 +1,25 @@
...
+
+$(basename $0) dev-\$FNHSNAME
+
+$(basename $0) local
 
 "
 
-ENVIRONMENT="${1:?"${USAGE}Please specify your environment name as the first parameter, e.g. dev-jane"}"
+ENVIRONMENT="${1:?"${USAGE}Please specify your environment name as the first parameter, e.g. dev-jane or local"}"
 WORKSPACE_TITLE="Selenium Testing"
```
alsuren, line 18: I ended up manually changing this to `Example workspace` to get ready for demo. I probably could have thought harder about how I was creating `Example workspace`, because now we have two competing workspaces that do different things in different ways, and neither of them have stable IDs that survive a reset-local.

alsuren: (I would be in favour of making this configurable or switching it to `Example workspace` by default.)

alsuren (approved):
Looks good. I would be a bit happier if we could port some of these scripts to typescript with generated graphql interfaces, but it looks like this wasn't as scary as I feared, so let's go with it.

frigus02:
I agree, that we should port this to TypeScript eventually. However making files work in the existing scripts was easier than this conversion, so I figured we might keep it as long as it works.

workspace-service/scripts/create-default-test-folders.sh
frigus02: I wouldn't mind changing this to "Example workspace". The individual scripts take the workspace name as a parameter. And this script could just set up this one workspace.

frigus02: I think I will do this in another PR, though. We might want to make this change also in the frontend stub server, Cypress and Selenium tests.

PR #434, opened 2020-10-29
Turn file url queries into mutations

frigus02:
We initially implemented both fileUploadUrls and fileDownloadUrl functions as GraphQL queries. Neither of them create any objects on the backend, so a query seemed more natural.

After using this for a bit, we noticed some issues:
- The React hook code generation assumes queries are called directly in   a components render function, whereas mutations are called based on   some user action, e.g. click. Calling queries based on a user action   leads to slightly awkward code, because we cannot use the generated  hooks.
- Queries are cached by the GraphQL caching middlewares by default,  unless they specify caching hints (maxAge or scope).

Thinking about this more, I can also find semantic reasons why those functions should be mutations. Both end up creating URLs, even if those are not stored on the backend. The URLs are meant to be used once by the client. In the future we might want to add tracking to file downloads, which actually does end up creating records on the backend, even if this is not something the user actively requests in the query.


StuartHarris:
This looks much better IMO, and I agree with your reasoning.

StuartHarris (approved):
Looks great, and works for me

PR #436, opened 2020-10-29
Show members UI

alsuren:
[Ticket]()

[Zeplin]()

## Pre-review checklist:

- Outstanding things:
  - [x] +/- for desktop view
  - [x] ```Warning: Each child in a list should have a unique "key" prop. Check the render method of `MobileList`. See https://fb.me/react-warning-keys for more information.```
  - [x] Everything marked FIXME in the code
  - [ ] ...?

image

- [ ] Tests

- [ ] Documentation

- [ ] Feature flag

## Testing information

- Test plan?

## Test:

- [x] Tested locally

- Platforms/Browsers:

  - [ ] Internet Explorer 11 - The icon rendering is a bit mental. Check that the active state doesn't suck on IE11.

  - [ ] Edge

- Accessibility:

  - [ ] Screen Reader

  - [ ] Keyboard Navigation

  - [ ] Magnification

  - [ ] Contrast

  - [ ] Text size 200%

  - [ ] Landscape (Mobile)


StuartHarris (approved):
reviewed together!

frontend/.env.development
```diff
@@ -1,3 +1,3 @@
 ORIGIN=http://localhost:3000
 COOKIE_SECRET=below-particular-sky-mix
-WORKSPACE_SERVICE_GRAPHQL_ENDPOINT=http://localhost:3001/graphql
+WORKSPACE_SERVICE_GRAPHQL_ENDPOINT=http://localhost:3030/graphql
```
frigus02, line 3: I think I would keep the stub server as the default, because it's the easier thing to use.

frontend/components/Expander/Expander.tsx
```diff
@@ -0,0 +1,47 @@
...
+`;
+
+interface Props {
+  expanded: boolean;
+  onClick: () => void;
+}
+
+const Expander: FC<Props> = ({ expanded, onClick }) => {
+  return (
+    <IconClickTarget onClick={onClick}>
```
frigus02, line 41: This might need some kind of alternative text on the icons or an aria label on the button.

frontend/pages/workspaces/[workspaceId]/members.tsx
```diff
@@ -29,12 +32,22 @@ const ContentWrapper = styled.div`
...
 `;
 
+const CountSentence = styled.p`
+  margin: 0;
+  padding-top: 16px;
+`;
+
+const nameCell: FC<User> = ({ name }) => <div>{name}</div>;
+const emailAddressCell: FC<User> = ({ emailAddress }) => (
+  <a href={`mailto:${emailAddress}`}>{emailAddress}</a>
```
Comment (frigus02) [inline frontend/pages/workspaces/[workspaceId]/members.tsx:42]: Do we need to escape `emailAddress` here, maybe with `encodeURIComponent`? I'm not entirely sure, to be honest. It can't break out of the `href` attribute, because React takes care of that. But would an email address like "jan\\?cc=foo@example.com" be possible?

Comment (frigus02) [reply frontend/pages/workspaces/[workspaceId]/members.tsx:42]: According to https://tools.ietf.org/html/rfc6068 the email address should be percent encoded:

> A number of characters that can appear in <addr-spec> MUST be percent-encoded.  These are the characters that cannot appear in a URI according to [STD66] as well as "%" (because it is used for percent-encoding) and all the characters in gen-delims except "@" and ":" (i.e., "/", "?", "#", "[", and "]").  Of the characters in sub-delims, at least the following also have to be percent- encoded: "&", ";", and "=".

I _think_ that encodeURIComponent should be fine, then.

```suggestion
  <a href={`mailto:${encodeURIComponent(emailAddress)}`}>{emailAddress}</a>
```

frigus02 (commented):
Looks good. Nice work 🙂

frontend/pages/workspaces/[workspaceId]/members.tsx
Comment (trcywu) [reply frontend/pages/workspaces/[workspaceId]/members.tsx:42]: React automatically encodes everything inside <href>

Comment (frigus02) [reply frontend/pages/workspaces/[workspaceId]/members.tsx:42]: Nevermind. `encodeURIComponent` encodes the "@" sign, which is not what we want. `encodeURI` could work.

frigus02 (approved):

declanslevin:
Something in the layout is pushing off screen when viewed on mobile: 
Screenshot 2020-11-05 at 15 08 55

It seems to be related to the padding-right of the PageContent container

In the interest of getting this PR merged soon, I'm happy to log this as a bug so as not to hold things up

trcywu:
yea please, we can pick it up separately as a bug. thanks!

declanslevin:
Logged under https://dev.azure.com/futurenhs/FutureNHS/_workitems/edit/2350

PR #450, opened 2020-11-02
Add analytics events

matt-paul:
[Ticket](https://dev.azure.com/futurenhs/FutureNHS/_boards/board/t/FutureNHS%20Team/Stories/?workitem=2245)

## Acceptance Criteria
This PR adds analytics events in where they don't exist already (generally due to workspace service functionality being created before events were up and running), as well as modifying some existing events with extra fields.

Existing events modified in this PR:
- [x] FileCreated - add 'versionNumber'
- [x] FolderCreated - add 'description'

New events added in this PR:
 - [x] FileDeleted
 - [x] FileDownload
 - [x] FolderUpdated
- [x] FolderDeleted

## Pre-review checklist:

- [ ] Tests

- [ ] Documentation

- [ ] Feature flag

## Testing information

- Test plan?

Backend only - Jamie to discuss with Declan about how to incorporate events into the QA process.

### DIFF
--- a/workspace-service/src/graphql/file_download_urls.rs
+++ b/workspace-service/src/graphql/file_download_urls.rs
@@ -15,7 +15,7 @@ impl FileDownloadUrlsMutation {
         let pool = context.data()?;
         let config = context.data()?;
         let event_client: &EventClient = context.data()?;
-        let requesting_user = context.data::<super::RequestingUser>()?;
+        let requesting_user: &super::RequestingUser = context.data()?;
         let user = db::UserRepo::find_by_auth_id(&requesting_user.auth_id, pool).await?;
         let id = Uuid::parse_str(&id)?;
         let file = db::FileWithVersionRepo::find_by_id(id, pool).await?;
--- a/workspace-service/src/graphql/file_download_urls.rs
+++ b/workspace-service/src/graphql/file_download_urls.rs
@@ -1,6 +1,7 @@
 use super::azure;
 use super::db;
 use async_graphql::{Context, FieldResult, Object, ID};
+use fnhs_event_models::{Event, EventClient, EventPublisher as _, FileDownloadedData};
 use url::Url;
 use uuid::Uuid;
 
@@ -13,8 +14,25 @@ impl FileDownloadUrlsMutation {
     async fn file_download_url(&self, context: &Context<'_>, id: ID) -> FieldResult<Url> {
         let pool = context.data()?;
         let config = context.data()?;
+        let event_client: &EventClient = context.data()?;
+        let requesting_user = context.data::<super::RequestingUser>()?;
+        let user = db::UserRepo::find_by_auth_id(&requesting_user.auth_id, pool).await?;
         let id = Uuid::parse_str(&id)?;
         let file = db::FileWithVersionRepo::find_by_id(id, pool).await?;
+        let folder = db::FolderRepo::find_by_id(file.folder, pool).await?;
+
+        event_client
+            .publish_events(&[Event::new(
+                id.to_string(),
+                FileDownloadedData {
+                    file_id: file.id.to_string(),
+                    user_id: user.id.to_string(),
+                    version_number: file.version_number.into(),
+                    version_id: file.version.to_string(),
+                    workspace_id: folder.workspace.to_string(),
+                },
+            )])
+            .await?;
 
         Ok(azure::create_download_sas(
             config,
--- a/workspace-service/src/graphql/file_download_urls.rs
+++ b/workspace-service/src/graphql/file_download_urls.rs
@@ -2,6 +2,7 @@ use super::azure;
 use super::db;
 use async_graphql::{Context, FieldResult, Object, ID};
 use fnhs_event_models::{Event, EventClient, EventPublisher as _, FileDownloadedData};
+use sqlx::PgPool;
 use url::Url;
 use uuid::Uuid;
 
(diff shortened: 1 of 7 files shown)


workspace-service/src/graphql/file_download_urls.rs
```diff
@@ -13,8 +14,25 @@ impl FileDownloadUrlsMutation {
     async fn file_download_url(&self, context: &Context<'_>, id: ID) -> FieldResult<Url> {
         let pool = context.data()?;
         let config = context.data()?;
+        let event_client: &EventClient = context.data()?;
+        let requesting_user = context.data::<super::RequestingUser>()?;
```
frigus02, line 18: For consistency you might want to put the type behind the variable everywhere.
```suggestion
        let requesting_user: &super::RequestingUser = context.data()?;
```

workspace-service/src/graphql/folders.rs
```diff
@@ -86,12 +88,14 @@ impl FoldersMutation {
         let pool = context.data()?;
         let workspace = Uuid::parse_str(&new_folder.workspace)?;
         let event_client: &EventClient = context.data()?;
+        let requesting_user = context.data::<super::RequestingUser>()?;
```
frigus02, line 91: Since you import `RequestingUser` above, you don't need `super::` here. You can also put the type behind the variable for consistency with the line above.

Thinking about it, I think the type annotation isn't needed at all. The type should be clear because we're passing `requesting_user` into `create_folder`. The same goes for the event client. I think you can remove the type annotation there, too.

```suggestion
        let requesting_user = context.data()?;
```

workspace-service/src/graphql/folders.rs
```diff
@@ -104,26 +108,29 @@ impl FoldersMutation {
         id: ID,
         folder: UpdateFolder,
     ) -> FieldResult<Folder> {
-        // TODO: Add event
         let pool = context.data()?;
-        let folder = db::FolderRepo::update(
-            Uuid::parse_str(&id)?,
+        let requesting_user = context.data::<super::RequestingUser>()?;
```
frigus02, line 112: Same as above.

workspace-service/src/graphql/folders.rs
```diff
@@ -132,41 +139,124 @@ async fn create_folder(
...
             },
         )])
         .await?;
     Ok(folder)
 }
 
+async fn update_folder(
+    id: ID,
+    title: &str,
+    description: &str,
```
frigus02, line 169: I think it would be nicer to accept `folder: UpdateFolder` here instead of title and description separately. When we use the `UpdateFolder` type, you can't accidentally mix up the order for the title and description arguments.

workspace-service/src/graphql/folders.rs
```diff
@@ -104,26 +108,29 @@ impl FoldersMutation {
...
-        Ok(folder.into())
+        .await
     }
 
     /// Delete folder (returns deleted folder
     async fn delete_folder(&self, context: &Context<'_>, id: ID) -> FieldResult<Folder> {
         // TODO: Add event
         let pool = context.data()?;
-        let folder = db::FolderRepo::delete(Uuid::parse_str(&id)?, pool).await?;
+        let requesting_user = context.data::<super::RequestingUser>()?;
```
frigus02, line 130: Same as above.

workspace-service/src/graphql/file_download_urls.rs
```diff
@@ -13,8 +14,25 @@ impl FileDownloadUrlsMutation {
...
+        let user = db::UserRepo::find_by_auth_id(&requesting_user.auth_id, pool).await?;
         let id = Uuid::parse_str(&id)?;
         let file = db::FileWithVersionRepo::find_by_id(id, pool).await?;
+        let folder = db::FolderRepo::find_by_id(file.folder, pool).await?;
+
+        event_client
+            .publish_events(&[Event::new(
+                id.to_string(),
+                FileDownloadedData {
```
frigus02, line 27: Do we want a test to check that this event is published correctly?

frigus02 (commented):
Looks good. Just a few small comments.

I do however have one bigger question. In this PR we're extending existing events with required fields. This means previously published events cannot be parsed using the new types. The only consumer for these events is analytics. So this is a question for Jamie. Is this going to be an issue in the data pipeline? If so, we can't extend the existing events and we need to create new versions instead.

workspace-service/src/graphql/folders.rs
```diff
@@ -132,41 +139,124 @@ async fn create_folder(
...
     workspace: Uuid,
     pool: &PgPool,
+    requesting_user: &RequestingUser,
     event_client: &EventClient,
 ) -> FieldResult<Folder> {
     let folder: Folder = db::FolderRepo::create(&title, &description, workspace, pool)
         .await?
         .into();
 
+    let user = db::UserRepo::find_by_auth_id(&requesting_user.auth_id, pool).await?;
```
alsuren, line 149: Probably not for this PR, but we have noticed that we need user.id a lot more often than we need user.auth_id. How do people feel about us sending down an x-user-id header, and making requesting_user have an id field?

frigus02: Yeah, I would prefer that. Then the only thing that has to deal with the auth_id is the `getOrCreateUser` function, which makes a lot of sense to me.

workspace-service/src/graphql/folders.rs
matt-paul: this is much cleaner now, ta!

workspace-service/src/graphql/folders.rs
matt-paul: ok, that makes sense

workspace-service/src/graphql/folders.rs
matt-paul: That sounds useful

alsuren (approved):
Rust code looks good. If Jan is happy with the answers to his comments then ship it.

frigus02 (approved):

matt-paul:
> Looks good. Just a few small comments.
> 
> I do however have one bigger question. In this PR we're extending existing events with required fields. This means previously published events cannot be parsed using the new types. The only consumer for these events is analytics. So this is a question for Jamie. Is this going to be an issue in the data pipeline? If so, we can't extend the existing events and we need to create new versions instead.

Chatted to Jamie - he has held off the pipeline work until this is through, so all good here.

PR #452, opened 2020-11-02
WorkspaceRepo[Fake]

alsuren:
This makes rust-analyser render things a bit better,
and gets us a little way towards having a better repo
pattern later.

https://docs.microsoft.com/en-us/dotnet/architecture/microservices/microservice-ddd-cqrs-patterns/infrastructure-persistence-layer-design#:~:text=of%20Work%20patterns.-,The%20Repository%20pattern,from%20the%20domain%20model%20layer. is a kind-of-reasonable description of the repository pattern, but we're not following it very accurately. We can iterate as we go.

### DIFF
--- a/workspace-service/src/db/mod.rs
+++ b/workspace-service/src/db/mod.rs
@@ -6,7 +6,12 @@ mod users;
 mod workspaces;
 
 pub use file_versions::*;
-pub use files::*;
+#[cfg(not(test))]
+pub use files::FileWithVersionRepo;
+#[cfg(test)]
+pub use files::FileWithVersionRepoFake as FileWithVersionRepo;
+pub use files::{CreateFileArgs, CreateFileVersionArgs, File, FileRepo, FileWithVersion};
+
 pub use folders::*;
 pub use teams::*;
 pub use users::*;
@@ -16,13 +21,9 @@ pub use workspaces::WorkspaceRepo;
 #[cfg(test)]
 pub use workspaces::WorkspaceRepoFake as WorkspaceRepo;
 
-#[cfg(not(test))]
-use {
-    anyhow::Result,
-    sqlx::{Executor, Postgres},
-};
+use anyhow::Result;
+use sqlx::{Executor, Postgres};
 
-#[cfg(not(test))]
 async fn defer_all_constraints<'c, E>(executor: E) -> Result<()>
 where
     E: Executor<'c, Database = Postgres>,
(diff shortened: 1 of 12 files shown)


workspace-service/src/db/mod.rs
```diff
@@ -6,18 +6,40 @@ mod users;
...
 #[cfg(not(test))]
+pub use users::UserRepo;
+#[cfg(test)]
+pub use users::UserRepoFake as UserRepo;
+
+pub use workspaces::Workspace;
+#[cfg(not(test))]
+pub use workspaces::WorkspaceRepo;
+#[cfg(test)]
+pub use workspaces::WorkspaceRepoFake as WorkspaceRepo;
```
alsuren, line 38: This conditional re-export dance is basically equivalent to Graeme's 
```
impl RealThing {}
impl MockThing {}
#[cfg(test)]
type Thing = MockThing;
#[cfg(not(test))]
type Thing = RealThing;
```
His may be clearer (or at least less deceptive). I can change it to that if you would prefer.

In practice, "go to definition" in vscode goes to the fake implementation in all cases.

workspace-service/src/db/workspaces.rs
```diff
@@ -15,13 +14,16 @@ pub struct Workspace {
...
+
+#[cfg_attr(test, allow(dead_code))]
+impl WorkspaceRepo {
     pub async fn create(title: &str, description: &str, pool: &PgPool) -> Result<Workspace> {
         let mut tx = pool.begin().await?;
 
-        let admins = Team::create(&format!("{} Admins", title), &mut tx).await?;
-        let members = Team::create(&format!("{} Members", title), &mut tx).await?;
+        let admins = db::TeamRepo::create(&format!("{} Admins", title), &mut tx).await?;
+        let members = db::TeamRepo::create(&format!("{} Members", title), &mut tx).await?;
```
alsuren, line 26: I would eventually like to be able to pass pool + repo objects in at the top-level (the graphql context) so that we can mock the repos out properly. The generic Executor thing makes this really painful.

The Microsoft article suggests only making Repositories at the roots of Aggregates. This may allow us to make the Repository use a &PgPool, and make everything that accepts an Executor be a private implementation detail. I think we need to sit down and draw aggregate boundaries around our models on https://miro.com/app/board/o9J_krvmqzA=/?moveToWidget=3074457350925392314&cot=12 , and I don't have the brain-space for that at the moment.

workspace-service/src/db/mod.rs
frigus02: This seems fine to me.

frigus02 (approved):

PR #454, opened 2020-11-03
Generate Rust event models with strings

frigus02:
I did a quick, timeboxed spike to generate Rust event models with ahead of time string building instead of build time macros.

Benefits:
- Run `yarn generate` to generate both TypeScript and Rust models. No need to manually edit any files.
- No Rust macros means faster builds and you can easily inspect the structs.
- We can easily add support for additional types such as `DateTime<Utc>`, which would have required us to fork `schemafy` before.

Downsides:
- Properly understanding everything of the JSON schema is hard. That's why the generation script doesn't even attempt to do this. It relies on a specific structure of the schema.json file.

I think I personally prefer this method. Interested to hear other opinions.


event-models/rust/src/models/generated.rs
```diff
@@ -0,0 +1,242 @@
...
+    /// The user that created the file
+    #[serde(rename = "userId")]
+    pub user_id: String,
+}
+
+///
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct FileUpdatedData {
+    ///
+    #[serde(rename = "fileId")]
```
alsuren, line 76: I was about to suggest `#[serde(rename_all = "camelCase")]` to reduce repetition, but actually I think your way is better because you don't have to agree with serde that `camelCase(toSnakeCase(name)) === name`.

event-models/rust/src/models/generated.rs
```diff
@@ -0,0 +1,242 @@
...
+    #[serde(rename = "versionId")]
+    pub version_id: String,
+
+    /// The user that created the file
+    #[serde(rename = "userId")]
+    pub user_id: String,
+}
+
+///
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
```
alsuren, line 73: Now I want to add `#[derive(schemars::JsonSchema)]` in a bunch of places, and see if we can make it round-trip and generate json schema for us ;-)

alsuren (approved):
I agree that this is a nicer approach than macro-based codegen. Happy to go with it once you've got your build passing.

Hopefully Matt doesn't get too tired of us refactoring things under his feet. 🤞

event-models/rust/src/models/generated.rs
frigus02: Yeah, `rename_all = "camelCase"` would definitely result in nicer code. But since this is auto-generated, I went with the safer option.

matt-paul:
I like this too, it's much clearer what is actually going on.

frigus02:
This changes the way the Rust code is generated, but it doesn't change the generated code itself. This is verified because it still compiles 🙂.

I'm going to mark this as no QA needed.

PR #464, opened 2020-11-05
Support fork PR builds

frigus02:
- Ticket: [AB#1673](https://dev.azure.com/futurenhs/75407963-f812-43e5-a734-1059bbc036a3/_workitems/edit/1673)

## Acceptance criteria

Builds triggered for PRs from forks should work 🙂 

Those builds don't have access to secrets. Luckily we're only using secrets to make it easy to test changes on a development environment in Azure. This is rarely necessary. Should be okay to skip for fork PRs.

Master builds still build and push images as normal.

## Testing information

Changes only affect the deployment process. See tests here:

- Successful tests for PR from fork: #460 
- Successful tests for PR from this repo: this one here. You're looking at it.

We only really test the master builds on merge.

### DIFF
--- a/.github/workflows/deploy-branch.sh
+++ b/.github/workflows/deploy-branch.sh
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-set -eux
+set -euxo pipefail
 
 COMPONENT=${1:?"Require components as first argument, e.g. frontend or workspace-service"}
 
--- a/.github/workflows/deploy-branch.sh
+++ b/.github/workflows/deploy-branch.sh
@@ -20,13 +20,13 @@ cd $HOME
 curl -s "https://raw.githubusercontent.com/kubernetes-sigs/kustomize/master/hack/install_kustomize.sh" | bash
 
 # Prepare manifests and copy to deployments repo
-cd $GITHUB_WORKSPACE/futurenhs-platform/$COMPONENT/manifests/dev-template
 if [ "$COMPONENT" = "infrastructure" ]; then
 	mkdir -p $GITHUB_WORKSPACE/futurenhs-deployments/cert-manager
 	cp -r $GITHUB_WORKSPACE/futurenhs-platform/infrastructure/kubernetes/cert-manager/* $GITHUB_WORKSPACE/futurenhs-deployments/cert-manager
 	mkdir -p $GITHUB_WORKSPACE/futurenhs-deployments/ingress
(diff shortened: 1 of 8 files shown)


.github/workflows/deploy-branch.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+
+set -eux
```
alsuren, line 3:
```suggestion
set -euxo pipefail
```
You probably want to be told when your curl call fails on line 20.

.github/workflows/deploy-branch.sh
```diff
@@ -0,0 +1,60 @@
...
+	$HOME/kustomize edit set image $DIGEST
+	$GITHUB_WORKSPACE/futurenhs-platform/infrastructure/scripts/create-dev-overlays.py
+	mkdir -p $GITHUB_WORKSPACE/futurenhs-deployments/$COMPONENT
+	cp -r $GITHUB_WORKSPACE/futurenhs-platform/$COMPONENT/manifests/* $GITHUB_WORKSPACE/futurenhs-deployments/$COMPONENT
+fi
+
+# Get commit message
+cd $GITHUB_WORKSPACE/futurenhs-platform
+git fetch origin HEAD --deepen=2
+COMMIT_MESSAGE=$(git log --no-merges -1 --format=%s)
```
alsuren, line 40: Thinking about it, I think that something like this might fix the commit messages:
```suggestion
cd $GITHUB_WORKSPACE/futurenhs-platform
git fetch origin $GITHUB_HEAD_REF --deepen=2
COMMIT_MESSAGE=$(git log origin/$GITHUB_HEAD_REF --no-merges -1 --format=%s)
```
(I haven't tried it out though, so might be best to do it in its own PR)

alsuren (approved):
Looks good. I would `-o pipefail` on all shellscripts. PRs with negative diffstat are my favourite kind.

.github/workflows/deploy-branch.sh
frigus02: The commit messages are actually fixed. I think I found the problem. We were using `${{ github.event.after }}` to refer to the commit. This properties does not exist in pull request events. Either it got removed at some point or maybe it only exists in push events. This new method works fine.

frigus02: As part of the build log you can see:

```
++ git log --no-merges -1 --format=%s
+ COMMIT_MESSAGE='Update master builds'
```

frigus02: Btw, you can see my attempts on the latest build on this PR: https://github.com/FutureNHS/futurenhs-platform/pull/463

alsuren (approved):

PR #470, opened 2020-11-06
Make Cypress workspace directory test reliable

frigus02:
This build failed because the Cypress test for the workspace directory page runs into a timeout: https://github.com/FutureNHS/futurenhs-platform/runs/1363111715

It waits for a page to load after click, which takes too long because it's compiling. Locally it generally works because our laptops are fast enough.

### DIFF
--- a/frontend/cypress/integration/frontend/workspaces/directory.js
+++ b/frontend/cypress/integration/frontend/workspaces/directory.js
@@ -9,13 +9,11 @@ describe("Workspace Directory", () => {
   it("Navigates to workspace page", () => {
     cy.visit(`/workspaces/directory`);
 
-    cy.get("div")
-      .contains(Cypress.env("TEST_WORKSPACE_NAME"))
-      .click()
-      .then(($h3) => {
-        const text = $h3.text();
-        cy.contains("h1", text);
-        cy.contains("h2", "Most recent items");
-      });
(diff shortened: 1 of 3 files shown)


frontend/cypress/integration/frontend/workspaces/directory.js
```diff
@@ -9,13 +9,11 @@ describe("Workspace Directory", () => {
...
-      .click()
-      .then(($h3) => {
-        const text = $h3.text();
-        cy.contains("h1", text);
-        cy.contains("h2", "Most recent items");
-      });
+    cy.get("div").contains(Cypress.env("TEST_WORKSPACE_NAME")).click();
+    cy.location("pathname", {
+      timeout: Cypress.config("pageLoadTimeout"),
+    }).should("eq", `/workspaces/${Cypress.env("TEST_WORKSPACE_ID")}`);
```
frigus02, line 15: Cypress has different default timeouts. It waits 60s for page loads. But it only waits 4s for commands. `cy.contains("h1")` is a command. So Cypress expects it to succeed in 4s. But sometimes compilation of the page in dev mode takes longer than that.

The fix is to wait for the page load timeout instead. I also added an assert to the URL, which seemed sensible.

declanslevin (approved):
Looks good and makes sense 🚢

PR #471, opened 2020-11-06
Make changes to create folder, Add workspace service mutations for create and update folder.

edcompton:
Add permissions to Create Folder page and implement edit and create workspace mutation changes. In the interest of being LEAN, I opted for using enums for graphQL -> Workspace service, then deserialising into strings once into the internals of the Workspace service.

- Add permissions radio buttons to create page.

- Edit Workspace service mutations for create.

- Edit Workspace service mutations for update.

In terms of design, no change from the 'Edit Folder' page that was reviewed previously. Only change is including the `Permissions` component. 

- Design: [Figma](https://www.figma.com/file/tYFsqhG61CleWS1tUapAar/Source-of-Truth-UI?node-id=685%3A337)

Screenshot 2020-11-09 at 12 05 26

## Pre-review checklist

- [ ] Tests (Coming in a later PR)
- [x] Documentation
- [ ] ~~Feature flag~~

## Testing information

Run the `frontend` and `workspace-service` locally using `yarn dev` and `make run-local` respectively.

- Test plan?

 Navigate to the create folder page and create a new folder with permissions. 

## Test checklist

- [x] Tested locally

- Platforms/Browsers:

  - [x] Internet Explorer 11
  - [x] Edge

- Accessibility:

  - [x] Screen Reader
  - [x] Keyboard Navigation
  - [x] Magnification
  - [x] Contrast
  - [x] Text size 200%
  - [x] Touch target areas (Mobile)
  - [x] Landscape (Mobile)


frontend/pages/workspaces/[workspaceId]/folders/[folderId]/update-folder.tsx
```diff
@@ -98,8 +102,23 @@ const UpdateFolder: NextPage = () => {
 
   const backToPreviousPage = () => router.back();
 
-  const onSubmit = async (result: any) => {
-    console.log(result);
+  const onSubmit = async (folder: Folder) => {
+    folder.id = folderId;
+    // Set the default folder permission to be all platform members
+    if (!folder.roleRequired) folder.roleRequired = "PLATFORM_MEMBER";
+    updateFolder({ ...folder }).then((result) => {
```
Comment (frigus02) [inline frontend/pages/workspaces/[workspaceId]/folders/[folderId]/update-folder.tsx:109]: Since this is an async function, you could use `await` here and avoid the nested `then` callback.

In any case it would be good to catch the promise rejection here, which I think can happen if the network call fails.

frontend/pages/workspaces/[workspaceId]/folders/[folderId]/update-folder.tsx
```diff
@@ -98,8 +102,23 @@ const UpdateFolder: NextPage = () => {
 
   const backToPreviousPage = () => router.back();
 
-  const onSubmit = async (result: any) => {
-    console.log(result);
+  const onSubmit = async (folder: Folder) => {
+    folder.id = folderId;
+    // Set the default folder permission to be all platform members
+    if (!folder.roleRequired) folder.roleRequired = "PLATFORM_MEMBER";
```
Comment (frigus02) [inline frontend/pages/workspaces/[workspaceId]/folders/[folderId]/update-folder.tsx:108]: Is `roleRequired` also set if the user never expanded the permission fields?

workspace-service/migrations/20201106130906_add_access_to_folders.sql
```diff
@@ -0,0 +1,4 @@
+-- Add migration script here
```
frigus02, line 1: You can remove this comment 🙂

workspace-service/src/db/folders.rs
```diff
@@ -99,6 +117,7 @@ impl FolderRepoFake {
             title: "fake folder".into(),
             workspace: Uuid::new_v4(),
             description: "fake folder for testing".into(),
+            role_required: "all-memebers".into(),
```
frigus02, line 120: Would that be `PLATFORM_MEMBER`?

workspace-service/src/db/folders.rs
```diff
@@ -124,6 +145,7 @@ impl FolderRepoFake {
             title: "fake folder".into(),
             workspace: Uuid::new_v4(),
             description: "fake folder for testing".into(),
+            role_required: "all-members".into(),
```
frigus02, line 148: Would that be `PLATFORM_MEMBER`?

workspace-service/src/db/folders.rs
edcompton: Yes, good catch

workspace-service/src/db/folders.rs
edcompton: same as above

frontend/pages/workspaces/[workspaceId]/folders/[folderId]/update-folder.tsx
Comment (edcompton) [reply frontend/pages/workspaces/[workspaceId]/folders/[folderId]/update-folder.tsx:108]: I wanted to set it so it always has a default value - no value is given to the outgoing data if the section is not expanded.

frontend/pages/workspaces/[workspaceId]/folders/[folderId]/update-folder.tsx
Comment (edcompton) [reply frontend/pages/workspaces/[workspaceId]/folders/[folderId]/update-folder.tsx:109]: Good point on both counts, was mirroring what was in the 'create folder' file but that's a better pattern.

frontend/pages/workspaces/[workspaceId]/folders/[folderId]/update-folder.tsx
Comment (frigus02) [reply frontend/pages/workspaces/[workspaceId]/folders/[folderId]/update-folder.tsx]: What does this mean for folder update? Will it always update the folder to `PLATFORM_MEMBER` unless you expand the section?

Comment (edcompton) [reply frontend/pages/workspaces/[workspaceId]/folders/[folderId]/update-folder.tsx]: Apologies, in my head we were looking at create - I see what you're angling at, you're right; this is an incorrect default as we don't want to edit permissions unless explicitly asked to by opening the section and selecting one of them. I've pushed a change.

frontend/pages/workspaces/[workspaceId]/folders/[folderId]/update-folder.tsx
```diff
@@ -98,8 +102,26 @@ const UpdateFolder: NextPage = () => {
 
   const backToPreviousPage = () => router.back();
 
-  const onSubmit = async (result: any) => {
-    console.log(result);
+  const onSubmit = async (folder: Folder) => {
```
Comment (frigus02) [inline frontend/pages/workspaces/[workspaceId]/folders/[folderId]/update-folder.tsx:105]: I think the type here is not correct. This should be `folder: Input`.

And then we might want to change the `Input` type to include the role field.

frontend/pages/workspaces/[workspaceId]/folders/[folderId]/update-folder.tsx
Comment (frigus02) [reply frontend/pages/workspaces/[workspaceId]/folders/[folderId]/update-folder.tsx]: Ah. You could make `role_required` in Rust an `Option<RoleRequired>`. Then we don't have to send it if the user didn't select anything.

frontend/pages/workspaces/[workspaceId]/folders/create-folder.tsx
```diff
@@ -69,19 +77,30 @@ const CreateFolder: NextPage = () => {
 
   const backToPreviousPage = () => router.back();
 
-  const onSubmit = async (newFolder: Folder) => {
-    createFolder({ ...newFolder, workspace: workspaceId }).then((result) => {
+  const onSubmit = async (newFolder: FolderInputs) => {
+    console.log(newFolder);
```
Comment (frigus02) [inline frontend/pages/workspaces/[workspaceId]/folders/create-folder.tsx:81]: The console.log can probably go.

frontend/pages/workspaces/[workspaceId]/folders/create-folder.tsx
```diff
@@ -129,13 +148,18 @@ const CreateFolder: NextPage = () => {
...
+            </FormField>
             <Button type="submit" name="submitButton">
               Save and complete
             </Button>
             <StyledButton secondary type="button" onClick={backToPreviousPage}>
               Discard
             </StyledButton>
-            {errors.server && <ErrorMessage>{errors.server}</ErrorMessage>}
+            {errors.server && (
+              <ErrorMessage>{errors.server.message}</ErrorMessage>
```
Comment (frigus02) [inline frontend/pages/workspaces/[workspaceId]/folders/create-folder.tsx:161]: I think we need to do the same change in update-folder.tsx

frigus02 (approved):

PR #494, opened 2020-11-10
Make Admin & Members buttons

trcywu:
image

- Ticket: AB#???
- Design: link to Figma

## Pre-review checklist

- [ ] Tests
- [ ] Documentation
- [ ] Feature flag

## Testing information

- Test plan?

## Test checklist

- [ ] Tested locally

- Platforms/Browsers:

  - [ ] Internet Explorer 11
  - [ ] Edge

- Accessibility:

  - [ ] Screen Reader
  - [ ] Keyboard Navigation
  - [ ] Magnification
  - [ ] Contrast
  - [ ] Text size 200%
  - [ ] Landscape (Mobile)

### DIFF
--- a/frontend/components/Table/Table.tsx
+++ b/frontend/components/Table/Table.tsx
@@ -30,6 +30,9 @@ const IconWrapper = styled.div`
 `;
 
 const StyledTable = styled(NHSTable)`
+  tbody tr:hover:not(&.withoutBottomBorder) {
+    background: ${({ theme }) => theme.colorNhsukWhite};
+  }
   td {
     font-size: 16px;
   }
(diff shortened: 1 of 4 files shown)


frontend/pages/workspaces/[workspaceId]/members.tsx
```diff
@@ -85,10 +89,32 @@ const WorkspaceMembersPage: NextPage = () => {
...
                       {
-                        heading: "Permissions",
+                        // eslint-disable-next-line react/display-name
+                        heading: () => <>Permissions</>,
                         // eslint-disable-next-line react/display-name
                         content: () => <>Administrator</>,
                       },
+                      {
+                        // eslint-disable-next-line react/display-name
+                        heading: () => <div>&nbsp;</div>,
```
Comment (alsuren) [inline frontend/pages/workspaces/[workspaceId]/members.tsx:99]: This prop was previously an optional string. Could we make the rendering for `heading: undefined` render like this, and keep this as a string?

Comment (trcywu) [reply frontend/pages/workspaces/[workspaceId]/members.tsx:99]: Setting `heading: undefined` renders the rows like this because the cell doesn't exist
image

I also tried an empty string which is the same effect. The best solution i could come up with was rendering an empty div.

frontend/pages/workspaces/[workspaceId]/members.tsx
```diff
@@ -104,10 +130,34 @@ const WorkspaceMembersPage: NextPage = () => {
...
+                          </>
+                        ),
+                      },
+                      {
+                        // eslint-disable-next-line react/display-name
+                        heading: () => <div>&nbsp;</div>,
+                        // eslint-disable-next-line react/display-name
+                        content: () => (
+                          <>
+                            <StyledButton secondary>
```
Comment (alsuren) [inline frontend/pages/workspaces/[workspaceId]/members.tsx:155]: I'm wondering if it would be valuable to make a component that renders the text and the button in a single cell. I suppose this could tie us in knots once we have the callbacks all hooked up though, so I'm happy to leave it like this until we have everything hooked up, and then see if there are any interesting refactoring opportunities.

Comment (trcywu) [reply frontend/pages/workspaces/[workspaceId]/members.tsx:155]: I think a lot will change between this PR and the end of the feature. Definitely an opportunity for a swoop of reactors at the end!

frontend/pages/workspaces/[workspaceId]/members.tsx
Comment (alsuren) [reply frontend/pages/workspaces/[workspaceId]/members.tsx:99]: I would prefer to keep the API of the Table component the same, and fix its implementation.

Comment (trcywu) [reply frontend/pages/workspaces/[workspaceId]/members.tsx:99]: aha. So apparently space in an empty string works.

frontend/components/Table/Table.tsx
```diff
@@ -143,29 +140,29 @@ const TableComponent = <ItemType extends Item>({
                     extraDetails.map((c, i) => {
                       const withoutBottomBorder = i < extraDetails.length - 1;
                       return (
-                        <NHSTable.Row key={i}>
-                          {c.heading && (
+                        <>
```
alsuren, line 143: feels like another unused fragment?

frontend/components/Table/Table.tsx
```diff
@@ -30,9 +30,6 @@ const IconWrapper = styled.div`
 `;
 
 const StyledTable = styled(NHSTable)`
-  tbody tr:hover {
-    background: ${({ theme }) => theme.colorNhsukWhite};
-  }
```
alsuren, line 35: I feel like this must have been there for a reason, but it's been the bane of our lives so maybe we cut our losses and kill it.

alsuren (approved):
remove the fragment and then go for it, assuming it renders okay (I've not checked since your recent pushes)

PR #496, opened 2020-11-11
Create 'get workspace permissions' query and frontend Error component. 

edcompton:
Screenshot 2020-11-11 at 09 00 23

## Acceptance criteria

- Workspace Service has a query that informs us of whether or not a given user is a member of a workspace.

- The frontend can render an error component when a user doesn't have sufficient permissions to access or edit a folder.

## Pre-review checklist

- [x] Tests
- [x] Documentation
- [ ] ~~Feature flag~~

## Testing information

- Test plan?

1. Change the permissions on a folder to be 'workspace members', by navigating to it's edit page and selecting 'Make visible to all members of this workspace' from the permissions section. Save the changes. grab the folder URL for later. 
2. Using tablePlus or similar, access the local DB.
3. Grab the ID of a user who belongs to the workspace you've just updated a folder in.
4. Run the local workspace service with `make run-local` and navigate to http://localhost:3030/graphiql.
5. Run the below query, using the workspace ID from the folder URL, and the user ID of the user from 2 steps above. In addition, add the following to the headers section in the bottom right `{ "x-user-auth-id": "feedface-0000-0000-0000-000000000000"}`.

```graphql
mutation {
  changeWorkspaceMembership(
    input: {
      workspace: [workspace id here]
      user: [user id here]
      newRole: NON_ADMIN
    }
  ) {
    id
    title
    admins: members(filter: ADMIN) {name, id}
    members: members(filter: NON_ADMIN) {name, id}
  }
}
```

6. In the frontend, logout with `localhost:3000/auth/logout`.
7. In the file `noop-passport-strategy.js`, change lines 15-17 to reflect the details of the user who's ID you used in the above query. 
8. Log back in with `localhost:3000/auth/login`. You are now authenticated with the user that has insufficient access permissions for the folder edited earlier.
9. Navigate to the folder URL from step 1. You should see an error.

## Test checklist

- [x] Tested locally

- Platforms/Browsers:

  - [x] Internet Explorer 11
  - [x] Edge

- Accessibility:

  - [x] Screen Reader
  - [x] Keyboard Navigation
  - [x] Magnification
  - [x] Contrast
  - [x] Text size 200%
  - [x] Touch target areas (Mobile)
  - [x] Landscape (Mobile)

### DIFF
--- a/frontend/lib/withUrqlClient.ts
+++ b/frontend/lib/withUrqlClient.ts
@@ -171,8 +171,16 @@ export default function withUrqlClient(
       if (process.env.NODE_ENV !== "production") {
         exchanges.unshift(devtoolsExchange);
       }
-
+      const fetchOptions = isServerSide
+        ? {
+            headers: {
+              // @ts-ignore
+              "x-user-auth-id": ctx?.req?.user.authId,
+            },
+          }
+        : undefined;
       return {
+        fetchOptions,
         exchanges,
         url: workspaceAPIServerUrl,
       };
--- a/frontend/pages/workspaces/[workspaceId]/folders/[folderId]/index.tsx
+++ b/frontend/pages/workspaces/[workspaceId]/folders/[folderId]/index.tsx
@@ -5,6 +5,7 @@ import Link from "next/link";
 import { useRouter } from "next/router";
 import styled from "styled-components";
 
+import { Error as ErrorComponent } from "../../../../../components/Error";
 import {
   IconCell,
   MobileModifiedAtCell,
@@ -26,9 +27,11 @@ import { PageLayout } from "../../../../../components/PageLayout";
 import { MobileList, Table } from "../../../../../components/Table";
 import {
   File,
+  RoleRequired,
   useFilesByFolderQuery,
   useGetFolderByIdQuery,
   useGetWorkspaceByIdQuery,
+  useGetWorkspaceMembershipQuery,
 } from "../../../../../lib/generated/graphql";
 import withUrqlClient from "../../../../../lib/withUrqlClient";
 
@@ -84,6 +87,23 @@ const FolderHomepage: NextPage = () => {
   const [files] = useFilesByFolderQuery({
     variables: { folder: folderId },
   });
+
+  const [returnedUserPermissons] = useGetWorkspaceMembershipQuery({
+    variables: {
+      workspaceId,
+    },
+  });
+
+  const userRole = returnedUserPermissons.data?.getWorkspaceMembership;
+  const folderAccessLevel = folder.data?.folder.roleRequired;
+
+  // If the folder only permits workspace members to view, and the user is not
+  // a member of the workspace, then set accessPermitted to false.
+  const accessPermitted =
+    folderAccessLevel === RoleRequired.WorkspaceMember &&
+    userRole === "NON_MEMBER"
+      ? false
+      : true;
 
   const items: MenuItem[] = [
     {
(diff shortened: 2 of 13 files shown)


frontend/lib/withUrqlClient.ts
```diff
@@ -171,8 +171,16 @@ export default function withUrqlClient(
       if (process.env.NODE_ENV !== "production") {
         exchanges.unshift(devtoolsExchange);
       }
-
+      const fetchOptions = isServerSide
+        ? {
+            headers: {
+              // @ts-ignore
```
matt-paul, line 177: What exactly are we ignoring here?

workspace-service/src/graphql/workspaces.rs
```diff
@@ -138,6 +154,25 @@ impl WorkspacesQuery {
...
+
+    // Checks workspace permissions for a given user ID and return the user if valid
+    async fn get_workspace_membership(
+        &self,
+        context: &Context<'_>,
+        workspace_id: ID,
+    ) -> FieldResult<NewRole> {
+        let requesting_user = context.data()?;
+        let pool = context.data()?;
+        let event_client: &EventClient = context.data()?;
```
matt-paul, line 166: Don't think we need this type

edcompton: good point

frontend/lib/withUrqlClient.ts
edcompton: props error: `Property 'user' does not exist on type 'IncomingMessage'.`.

matt-paul: ah, thats annoying.

frigus02: Yeah. We could possibly combine that with the check above. So that we're only ignoring once. E.g.

```ts
// @ts-ignore
const user: User | undefined = ctx.?req.?user;
if (!user && ctx.res) {
  ctx.res.writeHead(302, {
    Location: `/auth/login?next=${ctx.req.url}`,
  });
  ctx.res.end();
}
```

frontend/pages/workspaces/[workspaceId]/folders/[folderId]/update-folder.tsx
```diff
@@ -58,7 +61,11 @@ interface FolderInputs {
   server?: never;
 }
 
-const UpdateFolder: NextPage = () => {
+interface InitialProps {
+  user: User | undefined;
```
Comment (matt-paul) [inline frontend/pages/workspaces/[workspaceId]/folders/[folderId]/update-folder.tsx:65]: Do we need undefined here?

workspace-service/src/graphql/workspaces.rs
```diff
@@ -138,6 +154,25 @@ impl WorkspacesQuery {
         let workspace = WorkspaceRepo::find_by_id(id, pool).await?;
         Ok(workspace.into())
     }
+
+    // Checks workspace permissions for a given user ID and return the user if valid
+    async fn get_workspace_membership(
+        &self,
+        context: &Context<'_>,
+        workspace_id: ID,
+    ) -> FieldResult<NewRole> {
```
frigus02, line 163: I don't think `NewRole` makes sense as a return type here, because it's not a "new role". Maybe the easiest thing would be to rename `NewRole` to something like `WorkspaceMembership`.

frontend/pages/workspaces/[workspaceId]/folders/[folderId]/update-folder.tsx
Comment (edcompton) [reply frontend/pages/workspaces/[workspaceId]/folders/[folderId]/update-folder.tsx:65]: We actually don't need props at all. Removing the type.

workspace-service/src/db/workspaces.rs
```diff
@@ -123,6 +122,24 @@ impl WorkspaceRepo {
         }
     }
 
+    pub async fn is_member(workspace_id: Uuid, user_id: Uuid, pool: &PgPool) -> Result<Role> {
```
frigus02, line 125: Should we call thing something along the lines of `get_user_role`? I associate `bool` with functions called `is_xxx`.

workspace-service/src/db/workspaces.rs
```diff
@@ -123,6 +122,24 @@ impl WorkspaceRepo {
...
+        match db::UserRepo::find_by_id(&user_id, pool).await? {
+            Some(user) => {
+                let workspace = WorkspaceRepo::find_by_id(workspace_id, pool).await?;
+                let is_member = db::TeamRepo::is_member(workspace.members, user.id, pool).await?;
+                let is_admin = db::TeamRepo::is_member(workspace.admins, user.id, pool).await?;
+                if is_member {
+                    Ok(Role::NonAdmin)
+                } else if is_admin {
+                    Ok(Role::Admin)
```
frigus02, line 134: We need to swap those two around. Every admin is also a member. So we need to check for `is_admin` first. Otherwise this function never returns `Admin`.

workspace-service/src/graphql/workspaces.rs
```diff
@@ -107,6 +117,12 @@ struct UpdateWorkspace {
     description: String,
 }
 
+#[derive(InputObject)]
+struct GetMembership {
```
frigus02, line 121: Looks like this type is unused.

frontend/pages/workspaces/[workspaceId]/folders/[folderId]/index.tsx
```diff
@@ -85,6 +88,23 @@ const FolderHomepage: NextPage = () => {
     variables: { folder: folderId },
   });
 
+  const [returnedUserPermissons] = useGetWorkspaceMembershipQuery({
```
Comment (frigus02) [inline frontend/pages/workspaces/[workspaceId]/folders/[folderId]/index.tsx:91]: I think this way of checking permissions is not enough. The query `useGetFolderByIdQuery` must not return any folder information if the user doesn't have the permissions to see it. Otherwise we leak the folder data over the API.

workspace-service/src/graphql/workspaces.rs
edcompton: Makes sense to me.

workspace-service/src/db/workspaces.rs
edcompton: sounds good

frontend/pages/workspaces/[workspaceId]/folders/[folderId]/index.tsx
Comment (frigus02) [reply frontend/pages/workspaces/[workspaceId]/folders/[folderId]/index.tsx:91]: Sorry, that I didn't see this earlier.

I'm not sure if this query makes sense then. At least for this usecase, we could possibly look at the returned type/error of the `GetFolderById` query and use that to determine if we have access.

I'm not sure if `GetFolderById` should return a big enum with 2 cases:
- You have access (here is the folder data)
- You don't have access

Or if `﻿GetFolderById` should return an error. Apparently errors can be annotated with additional properties. So we don't have to make the decision based on the error message alone:
- https://async-graphql.github.io/async-graphql/en/error_handling.html
- https://async-graphql.github.io/async-graphql/en/error_extensions.html

workspace-service/src/graphql/workspaces.rs
```diff
@@ -253,6 +282,21 @@ async fn create_workspace(
     Ok(workspace)
 }
 
+pub async fn get_workspace_membership(
+    workspace_id: Uuid,
+    requesting_user: &RequestingUser,
+    pool: &PgPool,
+    _event_client: &EventClient,
```
frigus02, line 289: Hm, I guess we don't need the event client. Feel free to remove the parameter here.

workspace-service/src/db/workspaces.rs
frigus02: I think you only fixed this in the fake repo, not in the main one 🙂

workspace-service/src/graphql/workspaces.rs
edcompton: I've kept that in as think it's going to be used in my next PR when I tackle events.

frigus02 (approved):

PR #497, opened 2020-11-11
Record node error as attribute

frigus02:
The App Insights exporter doesn't export events at the moment. So this reports errors as attributes on the current span.


alsuren (approved):
looks good. If I go to http://localhost:3000/workspaces/invalidworkspaceid/download/invalidfileid then I get a log with `exception: '[GraphQL] invalid length: expected one of [36, 32], found 13',`

PR #502, opened 2020-11-12
Create sql query for file versions

matt-paul:
- Ticket: [AB#1805](https://dev.azure.com/futurenhs/FutureNHS/_boards/board/t/FutureNHS%20Team/Stories/?workitem=1805)

This micro PR has the sql query for file versions.  Given i'm now working on the frontend to stay out of the way of ongoing backend refactor, it would be good to get this merged in now so it's not sitting around locally.

### DIFF
--- a/workspace-service/sql/file_versions/find_by_file_id.sql
+++ b/workspace-service/sql/file_versions/find_by_file_id.sql
@@ -5,4 +5,4 @@ SELECT file_versions.id,
 
 FROM file_versions JOIN users ON file_versions.created_by = users.id
 WHERE file_versions.file = $1
-ORDER BY file_versions.created_at
+ORDER BY file_versions.version_number desc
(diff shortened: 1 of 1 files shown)


workspace-service/sql/file_versions/find_by_file_id.sql
```diff
@@ -0,0 +1,8 @@
+SELECT file_versions.id,
+    file_versions.file_title AS title,
+    file_versions.created_at,
+    users.email_address
+
+FROM file_versions JOIN users ON file_versions.created_by = users.id
+WHERE file_versions.file = $1
+ORDER BY file_versions.created_at
```
frigus02, line 8: Now that I think about it, you might want to order by `file_versions.version_number` instead. In theory this should be the same. But I don't trust dates. `version_number` is guaranteed to increase with every version.

Also, should this be ordered ascending or descending?

frigus02 (approved):

workspace-service/sql/file_versions/find_by_file_id.sql
matt-paul: Makes sense, i've also changed to descending - good prompt

PR #505, opened 2020-11-13
Hide folders from Nav bar when insufficient permissions.

edcompton:

## Acceptance criteria

- Cannot see folders that the logged in user can't access in the sidebar.
- Remove form 'flash' when loading the edit page with insufficient permissions.

### DIFF
--- a/workspace-service/src/graphql/folders.rs
+++ b/workspace-service/src/graphql/folders.rs
@@ -99,25 +99,14 @@ impl FoldersQuery {
         let requesting_user = context.data()?;
         let event_client: &EventClient = context.data()?;
         let folders = db::FolderRepo::find_by_workspace(workspace, pool).await?;
-        let first_folder = folders.first().unwrap();
-        let user_role = get_workspace_membership(
-            first_folder.workspace.clone(),
-            requesting_user,
-            pool,
-            event_client,
-        )
-        .await?;
+        let user_role =
+            get_workspace_membership(workspace, requesting_user, pool, event_client).await?;
         Ok(folders
             .into_iter()
             .map(Into::into)
             .filter(|folder: &Folder| {
-                if folder.role_required == RoleRequired::WorkspaceMember
-                    && user_role == WorkspaceMembership::NonMember
-                {
-                    false
-                } else {
-                    true
-                }
+                !(folder.role_required == RoleRequired::WorkspaceMember
+                    && user_role == WorkspaceMembership::NonMember)
             })
             .collect())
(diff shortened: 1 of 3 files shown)


workspace-service/src/graphql/folders.rs
```diff
@@ -96,8 +96,21 @@ impl FoldersQuery {
...
         let pool = context.data()?;
         let workspace = Uuid::parse_str(&workspace)?;
+        let requesting_user = context.data()?;
+        let event_client: &EventClient = context.data()?;
         let folders = db::FolderRepo::find_by_workspace(workspace, pool).await?;
-        Ok(folders.into_iter().map(Into::into).collect())
+        let first_folder = folders.first().unwrap();
+        let user_role =
+            get_workspace_membership(first_folder.workspace, requesting_user, pool, event_client)
+                .await?;
```
frigus02, line 105: It looks like you're using `first_folder` only to get the workspace ID. Unfortunately this will panic if there are no folders in a workspace.

I think instead of `first_wolder.workspace` you should be able to use the `workspace` variable directly. And then you don't need to panic, either :)

workspace-service/src/graphql/folders.rs
```diff
@@ -96,8 +96,21 @@ impl FoldersQuery {
...
+        Ok(folders
+            .into_iter()
+            .map(Into::into)
+            .filter(|folder: &Folder| {
+                !(folder.role_required == RoleRequired::WorkspaceMember
+                    && user_role == WorkspaceMembership::NonMember)
```
frigus02, line 111: I just realized. I think this does potentially exclude platform admins. We should probably do something like:

```suggestion
                !(folder.role_required == RoleRequired::WorkspaceMember
                    && user_role == WorkspaceMembership::NonMember
                    && !user_is_platform_admin)
```

workspace-service/src/graphql/folders.rs
edcompton: Great, thanks. Amazed I didn't spot that!

workspace-service/src/graphql/folders.rs
edcompton: Think this should be fixed by https://github.com/FutureNHS/futurenhs-platform/pull/504, which will declare any `platform_admin` as an Admin for any workspace.

frigus02: Ah brilliant.

frigus02 (approved):

PR #506, opened 2020-11-13
remove unused event_client arg; add tests

alsuren:
This is a follow on from https://github.com/FutureNHS/futurenhs-platform/pull/504 (which I cowboy-merged to unblock Tracy)


alsuren:
This will conflict with the huge refactor, and isn't super-valuable.

PR #507, opened 2020-11-13
Hexagonal workspaces

StuartHarris:
This PR refactors the `workspace-service` to fit the "ports-and-adapters" (or "hexagonal architecture") pattern in order to consolidate business logic in the core and make this easily testable.

See:
https://medium.com/better-programming/structuring-rust-project-for-testability-18207b5d0243
https://github.com/eckyputrady/testable-rust-web-project
http://www.dossier-andreas.net/software_architecture/ports_and_adapters.html

- [x] restructure into `core`, `services`, `repos`, `graphql` etc
- [x] stronger typing (using domain types)
- [x] add some more tests
- [ ] move all the existing tests
- [ ] make an events repo and add back the event sending
- [ ] rebase master on top

Raising this as a draft for now, so we can comment on the approach. I will add a diagram over the weekend to make it easier to understand.

PROS:
1. Much easier to write tests
2. Can see _all_ the business logic in one place (core + services) - keeping the adapters (repos and graphql) really thin.
3. Could easily add other adapters (e.g. a CLI adapter instead of the GraphQL adapter so we don't need scripts to e.g. set up a database).

CONS:
1. Lifetimes in Rust (I think I can collapse some of them so don't let them scare you off).
2. Everything needs to happen in the context of a DB transaction at the moment (although we can change this). It's not so bad, but probably not optimal.
3. More help from the type system (it's now impossible, for instance, to find a user by it's Id if you pass an auth Id.)


workspace-service/src/graphql/test_mocks.rs
```diff
@@ -37,7 +37,7 @@ pub async fn mock_admin_requesting_user() -> anyhow::Result<RequestingUser> {
 
 pub async fn mock_unprivileged_requesting_user() -> anyhow::Result<RequestingUser> {
     let pool = mock_connection_pool()?;
-    let requesting_user = db::UserRepo::get_or_create(
+    let requesting_user = db::UserRepoImpl::get_or_create(
```
alsuren, line 40: This feels like it shouldn't compile. I guess you commented out all of the `use crate::graphql::test_mocks::*;` calls?

workspace-service/src/graphql/workspaces.rs
```diff
@@ -156,17 +200,24 @@ impl WorkspacesQuery {
...
-        requesting_user_workspace_rights(
-            workspace_id.try_into()?,
-            requesting_user,
-            pool,
-            event_client,
-        )
-        .await
+        let workspace_service = context.data::<WorkspaceServiceImpl>()?;
+        let pool: &PgPool = context.data()?;
+        let mut repos = RepoFactory::new(pool.begin().await?);
```
alsuren, line 205: We could probably collapse these two lines into one if we wrote an extension trait on Context

```
let mut repos = context.repos().await?;
```

Want me to have a stab at this or is it too magical?

workspace-service/src/graphql/workspaces.rs
```diff
@@ -156,17 +200,24 @@ impl WorkspacesQuery {
...
+
+        let membership = workspace_service
+            .requesting_user_workspace_rights(
+                &mut repos,
+                workspace_id.into(),
+                requesting_user.auth_id.into(),
+            )
+            .await?;
+
+        repos.commit().await?;
```
alsuren, line 218: This feels like a footgun. Could we have a Drop impl on repos that logs something if it's not been committed or something? (not that anyone would notice, because the workspace service is quite spammy)

I think what I would really like would be if there was a hook in async-graphql where we autocommit on success and auto-rollback on failure or something.

That said, maybe this is the best approach, as long as we apply it everywhere.

workspace-service/src/core/workspace.rs
```diff
@@ -0,0 +1,157 @@
...
+    where
+        T: RepoCreator<'b> + Send,
+        'b: 'a;
+
+    async fn create_workspace<T>(
+        &self,
+        repo_factory: &'a mut T,
+        title: &str,
+        description: &str,
+        requesting_user: AuthId,
```
alsuren, line 108: (not worth blocking the PR for, but it's worth mentioning) In my head, the `RequestingUser` struct is a domain object that's worth passing around as a struct. I would eventually like to add an x-user-id header instead of x-user-auth-id, to make our lives easier.

workspace-service/src/graphql/workspaces.rs
alsuren: Thinking about it further, I think my concern is that this is this is halfway between "the web framework manages the transactions" and "the services manage the transactions" (the latter would allow you to write a repofactory impl for tests that can be told whether or not to expect a commit). We will want to migrate in one direction or the other eventually, but this is probably okay for now.

